PR #2, opened 2026-06-12
Native subclassing, WeakRef correctness, and conformance tooling

mparrett:
# Native subclassing, WeakRef correctness, and conformance tooling

A conformance-focused branch that tightens places where the runtime model was too
approximate for ECMAScript semantics, plus the test262 tooling used to measure it.

## Conformance

- **TypeScript checker (tests/cases/conformance/classes):** 55.5% → **66.0%** (+48 tests),
  from strict property initialization (TS2564) and definite-assignment assertions.
- **test262 built-ins:** ~72%, flat within 0.2s-timeout noise (the suite is timing-sensitive at
  that bound). The built-ins changes here are correctness, not pass-count: WeakRef construction
  now runs (measured +10), and subclassed Array/Map/Set/Iterator get correct prototypes
  (previously they panicked or were silently mis-typed). Language suite is ~flat.

## Changes

**Native subclassing (`class S extends Array {}` and friends)**
- Thread `new.target` and `GetPrototypeFromConstructor` through `OpNew`, `OpSpreadNew`, and
  the native constructor branches, so Array/Map/Set/Iterator/WeakRef produce instances with
  the subclass prototype.
- Give Array/Map/Set per-instance prototype overrides, so an exotic builtin instance still
  gets ordinary prototype-chain behavior — including **subclass method overrides** of builtin
  names (`push`/`set`/`add`), which previously lost to the intrinsic fast path.

**Runtime correctness**
- `WeakRef.Deref()` preserves the target's VM type instead of rewrapping everything as a
  plain object — matters for arrays/functions and any type-specific op after deref.
- Make the VM cancellation flag atomic, closing a data race between the interpreter loop and a
  cancelling goroutine.

**Checker**
- Strict property initialization (TS2564) and definite-assignment assertions, gated on
  `@strict` / `@strictPropertyInitialization` directives.

**test262 tooling**
- JSON output, failure clustering, allow-list mode, stdout isolation, timeout cancellation,
  baseline diffing, and a pinned test262 revision — so conformance work is measurable without
  fighting the harness.

**Cleanup**
- Remove two committed backup files (`pkg/parser/parser.go.bak`, `pkg/vm/vm.go.orig`, ~19k
  lines). This accounts for nearly all of the diff's deletions; the code change is ~+1,164 lines.

## Testing

`go build ./...` and `go test ./tests -run TestScripts` green. Adds runtime smoke tests for
subclassed natives (including builtin-method overrides) and WeakRef deref.

## Follow-ups

Tracked separately, not in this PR:

- The test262 runner can call cleanup/reset while a timed-out VM goroutine is still executing
  — a flag-race-adjacent hazard the atomic change doesn't fully cover.
- Strict-property-init uses an any-path approximation where TypeScript does full control-flow
  analysis (e.g. `constructor(c) { if (c) this.x = 1 }` isn't flagged). A precision gap, not a
  crash.

---

Developed with help from Claude Opus 4.8 and Codex; per-commit AI co-authorship is in the trailers.


mparrett:
@nooga I ran this as an extension of your original AI speedrun, mostly to see if the bots could easily extract some wins. I don't have much skin in the game, but figured you might like to see it at the very least!

nooga:
Hi @mparrett thanks so much! I have a huge portion of compat work sitting in my local checkout at the moment, I haven't pushed this for a while so this might become a hardcore three-way merge! 

This engine is actually useful even as is (I'm using it in prod for some less critical tasks) but I want to bring it to 90%+ compat with TS before launching it :D 

I'll look into your PR and see how to merge it cleanly 🙏

mparrett:
I thought the below map might help if you decide to try that hairy merge.

# Change map: `feature/conformance-and-vm-fixes`

A stage-by-stage map of the 44 commits, oriented for reconciliation rather than
review. Each stage gives the intent (what/why) and the general approach, then
lists the commits and the subsystems it touches — so you can quickly see where it
overlaps your unpushed work. The commit bodies have the line-level detail; this is
the altitude above them.

Reading order is roughly chronological, with the docs-tree churn collected into
one stage so it doesn't fragment the engine narrative.

---

## 1. testtsc baseline matching

**What/why:** the TS conformance harness mis-resolved parameterized-variant
baselines, so multi-baseline tests were scored against the wrong expected output.
**How:** match a test's parameterized baselines against its `// @` directives, and
ignore variants with no source errors.
**Commits:** `8412d33`, `096cd1e`, `a9cbec2`
**Touches:** `cmd/paserati-testtsc`. Measurement only — no engine change.

## 2. TS2564 strict property initialization

**What/why:** add the checker diagnostic for class fields that aren't definitely
assigned, plus definite-assignment assertions. A conformance gain
(classes 55.5% → 66.0%).
**How:** walk the constructor for `this.x = …` assignments and flag fields never
seen. It's an **any-path approximation** (init on any branch counts), not TS's
full control-flow analysis — a known precision gap. Gated on
`@strict` / `@strictPropertyInitialization`; the no-directive default is **on**,
which is deliberate: we measured that flipping it off regresses classes 66.0% →
58.8% because the conformance baselines run strict effectively on.
**Commits:** `f2b6363`, `1d5d037`, `9eb17e1`
**Touches:** `pkg/checker` (`strict_init.go`), `cmd/paserati-testtsc` gating.

## 3. Variadic double-diagnostic fix

**What/why:** `checkCallExpression` visited variadic-position arguments twice, so
errors inside them (e.g. `console.log(undefinedVar)`) emitted duplicated.
**How:** let the variadic loop own those arguments; drop the redundant catch-all
visit and the fixed-args gate that was silently swallowing variadic errors.
**Commits:** `bc85d9e`
**Touches:** `pkg/checker` (call checking).

## 4. Test262 harness: reproducibility and tooling

**What/why:** make conformance reproducible from a clean checkout and the harness
usable for gating, rather than a one-off script.
**How:** pin the Test262 revision (`.test262-rev`) with a non-interactive setup;
extract the shared `Stats`/`Result` types into a new **`pkg/test262`** used by both
the runner and analyze; make `-json` a clean contract (swap `os.Stdout`→stderr
during test exec so `console.log` can't corrupt the JSON, restore for the final
encode); add `-allow-failures <baseline>` (exit 0 iff every current failure is in
the baseline) to gate regressions without a fully-green suite; fail on timeouts;
tag-based skips for network/slow/flaky scripts; anchor `.gitignore` so `test262/`
stops matching `pkg/test262/`.
**Commits:** `712f454` (drop tracked `.bak`/`.orig`), `d67511d`, `d7f9965`,
`ad5be3b`, `fed874c`, `ee987c6`, `35871d0`
**Touches:** `cmd/paserati-test262`, `cmd/paserati-analyze`, new `pkg/test262`,
`setup-test262.sh`, `.gitignore`. **The `pkg/test262` extraction is the most
likely structural conflict with unpushed runner changes.**

## 5. Docs-tree reorganization

**What/why:** internal cleanup: condensed `AGENTS.md` (624→166 lines), moved
completed design/postmortem docs into a new `docs/archived/`, extracted remaining
open work into tickets, and scrubbed hard-coded pass-rate snapshots that had gone
stale. **This relocates a lot of your `docs/*.md` and rewrites `AGENTS.md`**, so
it's the stage most likely to be unwanted or to collide with your tree — easy to
drop or cherry-pick independently of the engine work.
**Commits:** `67746b5`, `fea74ac`, `57898b4`, `92bcf24`, `84cdf72`, `bbda723`,
`97323ab`, `85d309e`, `0483967`, `2fea112`, `05f4785`, `0e4f9a7`, `e4ac5b1`,
`373765e`, `7db550a`, `f573aad`, `8850c3e`, `69c8cd2`, `24c7bc0`, `f656a38`
**Touches:** `docs/`, `AGENTS.md`. No engine code.

## 6. VM correctness and a slow-path perf fix

**What/why:** three independent runtime fixes.
- **WeakRef deref type laundering** (`a3f9582`): `Deref()` returned everything as a
  plain object regardless of the stored target, so a deref'd array/function lost
  its VM type. Now preserves the target's type.
- **Cancellation data race** (`6eeaeaf`): the VM `cancelled` flag was a plain bool
  written by the timeout watchdog and read by the interpreter loop with no
  synchronization. Switched to `atomic.Bool`.
- **O(1) slow-path property lookup** (`7875275`): `PlainObject.GetOwn` scanned
  `shape.fields` linearly on every cache miss. Added a lazy, atomically-published
  per-Shape name→index map above an 8-field threshold (shapes are append-only, so
  it never goes stale). ~5× on large shapes.
**Commits:** `a3f9582`, `6eeaeaf`, `7875275`
**Touches:** `pkg/vm` — WeakRef, the interpreter cancellation path, and **`Shape`**
(the perf change is the one to diff against any unpushed Shape/IC work).

## 7. Native subclassing

**What/why:** make `class S extends Array / Map / Set / Iterator / WeakRef {}`
work — instances carry the subclass prototype, `instanceof S` is true,
and subclass method overrides win. Pre-fix, derived instances got the intrinsic
prototype (or the constructor silently exited via an `OpSpreadNew` type-pun panic).
**How (the through-line):**
- Thread `new.target` through `OpNew`'s native-ctor branches (and the matching
  `OpSpreadNew`/bound-fn sites), mirroring the `Reflect.construct` path — native
  ctors were setting `inConstructorCall` but never `currentNewTarget`, so
  `new WeakRef(x)` errored "requires 'new'" (`b63ef93`).
- Give Array/Map/Set a per-instance `prototype` override; `op_getprop` and
  `OpInstanceof` consult it before the intrinsic. Defaults unset, so plain `[]` /
  `new Map()` are unchanged. Propagate `newTarget` through `super()` into native
  ctors now that prototype threading makes it safe (`44264c2`).
- Thread `newTarget` + `GetPrototypeFromConstructor` into the Array/Map/Set/
  Iterator constructors themselves so they pick the subclass prototype (`a73eb80`).
- Route `TypeWeakRef` property access through `WeakRef.prototype` (it had no branch,
  so `wr.deref` came back undefined even after construction worked) (`0850611`);
  accept callable targets in the `WeakRef` ctor (`0623e49`).
- Final correctness fix from external review: subclass **overrides of builtin
  method names** (`push`/`set`/`add`) were still losing to the intrinsic fast path
  in `handlePrimitiveMethod`, which consulted the realm intrinsic before the
  per-instance prototype. Now it consults the instance prototype first (`c44fe94`).
**Commits:** `b63ef93`, `0850611`, `0623e49`, `44264c2`, `a73eb80`, `19b7d7d`
(subclass smoke + language-baseline regen), `c44fe94`
**Touches:** `pkg/vm` — `OpNew`/`OpSpreadNew`, `op_getprop`, `OpInstanceof`,
`handlePrimitiveMethod`, the `ArrayObject`/`MapObject`/`SetObject` representations,
and `pkg/builtins` Array/Map/Set/Iterator/WeakRef constructors. **This is the
deepest engine change and the most reconciliation-sensitive — newTarget plumbing,
prototype lookup order, and constructor-state restore all touch broad surfaces.**

---

## Reconcile checklist

- **Engine, high overlap risk:** stages 6–7 (`pkg/vm`: newTarget, per-instance
  prototypes, prototype lookup order, Shape index).
- **Tooling, structural overlap risk:** stage 4 (`pkg/test262` extraction).
- **Independent / droppable:** stage 5 (docs + `AGENTS.md`) — separable from
  everything else.
- **Low-risk additions:** stages 1–3 (checker + testtsc).

nooga:
Thanks for the contribution! This is a useful set of fixes: native subclassing, `WeakRef` behavior, stricter property initialization, and the conformance/tooling cleanup all move the runtime in the right direction.

I rebased this onto latest main and fixed a few follow-up edge cases while I was there: subclass `Map`/`Set` symbol overrides now win, `WeakRef` subclasses work with `instanceof`, and the `in` operator sees subclass prototype methods on native instances. I also adjusted the narrowing smoke test to be explicit about definite assignment so it still tests narrowing rather than initialization.

Validated with `go build ./...`, `go test ./tests -run TestScripts`, pinned Test262 checks, and the TS conformance suite. No deterministic regressions found; the only noisy diffs were in already-flaky staging/sm-style runs.

PR #4, opened 2026-06-26
perf: bench-ratchet + same-runner A/B PR check

mparrett:
## What
A perf-tracking check for PRs, built to be honest about GitHub's shared runners.

`cmd/bench-ratchet` runs the Go benchmarks and normalizes each against a frozen calibration anchor (`BenchmarkRatchetAnchor`, a pure-CPU FMA loop), so a benchmark reads as a multiple of the anchor rather than an absolute time.

## The PR check (`perf-pr.yml`)
Comparing against a *stored* baseline doesn't hold up on shared runners: they vary by CPU tier, and an ALU anchor and a pointer-chasing lookup don't scale the same across CPUs, so the ratios drift ~20% run to run. So the check is a **same-runner A/B** that benches the PR's merge-base and head on one runner, back to back, and posts the delta. Running both halves on one machine cancels the cross-runner noise.

- **Informational: it never fails the PR.** The micros have real within-run variance, so it reports deltas for a human to read, not a gate.
- **Opt-in via the `perf` label**, since a two-pass bench costs runner minutes. Add the label to run it; remove it to stop.
- Posts to the run summary and a sticky PR comment. The summary is the channel that works on fork PRs, where the token is read-only.

Mirrors `let-go`'s `perf-pr.yml`.

## Notes
First of two PRs; #5 adds the benchmark-trend timeline page on top of this. Adds one dependency, `golang.org/x/perf` (the Go team's `benchfmt` parser).


mparrett:
Self-review found one issue worth fixing before merge:

`cmd/bench-ratchet/main.go:520-523` downgrades benchmark package failures to warnings and continues. The missing entries are later reported by the compare path, but the command still exits 0 unless there is a measured regression. I reproduced this by running an anchor-only `check` against the committed baseline: it reported `30 missing` benchmarks and exited 0.

That means if `tests` benchmarks stop compiling, time out, or are accidentally renamed/removed, the PR gate can pass with much less benchmark coverage. For CI, package capture errors and/or missing baseline entries should fail `check` unless an explicit opt-out flag is used.

mparrett:
Fixed in 0e5d632. `check` now fails when any package capture errors or any baseline benchmark is missing, with `-allow-incomplete` to opt out. Reproduced the anchor-only case — it exits 1 with `30 missing benchmark(s)` now instead of 0, and full-coverage runs still pass.

nooga (approved):

PR #5, opened 2026-06-26
perf: benchmark trend page ("are we fast yet?")

mparrett:
## What
A static "are we fast yet?" page that charts benchmark performance over time, built on the ratchet from #4, and roughly following the same pattern from `let-go`. Live preview on my fork (screenshot below): https://mparrett.github.io/paserati/

## How it works
- `perf-timeline.yml`: on each push to `main`, snapshots the benchmarks and pushes one immutable JSON to a dedicated `perf-data` orphan branch. Keeping it off `main` keeps the data append-only and sidesteps branch protection.
- `pages.yml`: pulls those snapshots and deploys `docs/perf/index.html` to GitHub Pages, auto-redeploying after each snapshot.
- The page plots each benchmark's ratio-to-anchor over time, normalized to its baseline (1.0× is the bar). Two views: **Overall** (mean with a ±1σ band) and **Breakdown** (per-benchmark). It renders client-side and runs no benchmarks.

## Setup required
- Enable GitHub Pages: Settings → Pages → Source = GitHub Actions. The page shows an empty state until the first snapshot lands.
- The `perf-data` branch auto-creates on the first timeline run; it's unprotected and holds only snapshot JSON.

## Notes
Stacked on #4 — this diff includes the gate until #4 lands. The mascot sticker is the hero; swap or drop it freely. Forward-only: it tracks new commits as they land (backfilling old commits is possible but out of scope).

---

image


mparrett:
Self-review found one PR-specific workflow issue:

`actionlint` rejects `.github/workflows/perf-timeline.yml:32`:

```text
unexpected key "queue" for "concurrency" section. expected one of "cancel-in-progress", "group"
```

That means the workflow will fail syntax validation before it can snapshot benchmarks. We should drop `queue: max` and rely on `cancel-in-progress: false` / default queuing behavior, or use a supported serialization pattern.

Final checks also reran `go test ./tests/...` successfully. `actionlint` reports an existing unrelated warning in `.github/workflows/release.yml`; I did not count that against this PR.

mparrett:
Fixed in 3ae4d26. Dropped `queue: max`: the timeline is forward-only (one snapshot per push), so there are no bursts to queue, and `cancel-in-progress: false` already serializes the perf-data writers. `actionlint` is clean now — it also flagged an SC2012 `ls` in `pages.yml`, fixed in b448907.

nooga (approved):

nooga (approved):

PR #10, opened 2026-06-28
perf: Test262 macro-benchmark (opt-in same-runner A/B)

mparrett:
Follows the perf work in #4 (`bench-ratchet`) and #5 (timeline page): the suggested **"total Test262 run time"** macro-benchmark.

## What it is

An **opt-in, same-runner A/B** that answers "did this PR make executing the Test262 corpus faster or slower?" It benches the PR's merge-base and head **back-to-back on one runner** and reports the delta. Informational — **it never fails the PR**.

- **Metric:** sum of per-test execution time over tests passing (and not timing out) on **both** sides. Pass/fail/timeout counts are reported **separately**, so a correctness change moves those numbers, not the timing.
- **Opt-in:** runs only on PRs labeled **`perf-test262`** (distinct from the cheap `perf` micro-check — adding `perf` must never kick off this multi-minute job), plus `workflow_dispatch`.
- **Scope:** `built-ins` + `language` (the JS-execution core).

## Why these choices (each was validated, several reversed a first guess)

- **Raw summed-ns, not anchor-normalized.** A same-runner A/B has no cross-runner drift to cancel; normalizing each half by its own separately-measured anchor only *injects* the anchor's measurement noise. (The anchor stays where it belongs — `bench-ratchet`.)
- **Sharded, one process per second-level dir.** A single process over the whole corpus grows VM state unbounded and OOMs a 7 GB runner (**5.3 GB → 439 MB peak** after sharding). Per-test durations are process-independent, so this doesn't change the metric.
- **1.2s per-test timeout.** 5s makes the suite timeout-dominated (40+ min); 0.2s mislabels slow-but-correct tests (e.g. RegExp backtracking) as timeouts and drops them. Every *measured* slow-correct test resolves under 1.2s; what still times out is a genuine hang.
- **Intersection compare.** Summing each side's passing set independently let a boundary test flipping pass↔timeout between passes move the delta. Comparing only tests passing on both sides drops identical-code noise from +1.57% to **−0.23% total**.

## Validation (CI self-A/B on identical code via `workflow_dispatch` `ref=`)

| | |
|---|---|
| Full A/B (2 passes + setup) | ~6.5 min |
| Peak RSS | 439 MB |
| **Total noise floor (identical code)** | **−0.23%** |
| Correctness | 40,194 passed / 7,085 failed / 12 timeouts, stable across passes |

**Caveat:** per-suite deltas swing ~±3% on identical code (shared-runner CPU contention across each ~3-min pass; the suites nearly cancel in the total). **Read the `total`** (sub-1% noise); per-suite is diagnostic with a ~±4% band.

## To use after merge

Add the **`perf-test262`** label to a PR (one-time label creation), or `workflow_dispatch` with a `pr` number. To smoke-test the harness itself, `workflow_dispatch` with a `ref` (runs base == head, expect ~0%).

## Notes

- New files only: `cmd/bench-test262/`, `scripts/macro-test262*.sh`, `.github/workflows/perf-test262.yml` (+ two `.gitignore` lines). No existing code touched.
- `go build ./...`, `go vet`, and `go test ./tests -run TestScripts` are green.
- Like `perf-pr.yml`, it's fork-safe (job-summary + best-effort sticky comment; a read-only fork token degrades gracefully).

### DIFF
--- a/.github/workflows/perf-test262.yml
+++ b/.github/workflows/perf-test262.yml
@@ -1,0 +1,213 @@
+name: Perf Test262
+
+# Same-runner A/B macro-benchmark: how long paserati takes to execute the whole
+# Test262 corpus. Benches the PR's merge-base and its head back to back on ONE
+# runner, then reports the delta. Like perf-pr.yml, running both halves on the
+# same machine cancels cross-runner CPU-tier noise.
+#
+# The metric is the SUM of per-test execution time over the tests that pass and
+# don't time out, divided by the calibration anchor (anchor-normalized, machine-
+# portable). Summing per-test durations (not wall-clock) keeps it parallelism-
+# invariant; restricting to the passing, non-timed-out set decouples speed from
+# correctness. Pass/fail/timeout counts are reported separately, so a correctness
+# change moves those numbers, not the timing.
+#
+# Heavy and opt-in: a two-pass full-suite run costs real runner minutes, so it
+# runs only on PRs carrying the `perf-test262` label (distinct from the cheap
+# `perf` micro check — adding `perf` must never trigger this multi-minute job).
+# Informational only: it never fails the PR.
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+    paths:
+      - 'pkg/**'
+      - 'cmd/**'
+      - '.test262-rev'
+      - '.github/workflows/perf-test262.yml'
+  # Manual escape hatch: run against any PR by number without touching labels.
+  workflow_dispatch:
+    inputs:
+      pr:
+        description: 'PR number for a real base-vs-head A/B'
+        required: false
+      ref:
+        description: 'Single ref to self-A/B (smoke test: runs the macro twice, expect ~0% — measures wall-time + noise floor)'
+        required: false
+
+permissions:
+  contents: read
+
+concurrency:
+  group: perf-test262-${{ github.event.pull_request.number || github.event.inputs.pr || github.event.inputs.ref }}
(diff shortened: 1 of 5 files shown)


nooga (approved):

mparrett:
Retrospective review: I found three issues that remain relevant on main and to the related timeline work.

**[P1] Pin the same Test262 corpus for both halves**

Each checkout runs setup-test262.sh using that revision of .test262-rev. A PR changing the pin therefore compares different test source under matching paths, and the intersection comparator reports corpus changes as engine performance changes. This is especially problematic because .test262-rev explicitly triggers the workflow. Please resolve one corpus revision, normally the head revision, and use it for both halves, or refuse to report a performance delta when the pins differ.

**[P2] Preserve the contributing test set in aggregate output**

cmd/bench-test262 emits only the sum and count of the current run passing tests. It loses test identity, so two records with equal counts can still represent different workloads. The A/B comparator avoids this by reading the per-test merged files, but the aggregate output advertised for timeline use cannot; PR #16 exposes exactly this problem. Please include a stable-set fingerprint or per-test data, or do not expose this total as a comparable series.

**[P2] Allow harness-only PRs to trigger the opt-in workflow**

GitHub applies the pull_request paths filter before the labeled job condition. Adding perf-test262 to a PR that only changes macro-test262.sh, macro-test262-compare.sh, or setup-test262.sh therefore does not run the workflow. Since the label is already an explicit heavy-test opt-in, please remove the paths filter or include every harness and dependency file.

The workflow passes actionlint. I did not rerun the full Test262 macro.

PR #13, opened 2026-06-28
checker: use destructuring annotations for binding types

nooga:
## Summary
- Contextually type annotated array destructuring initializers when the RHS is an array literal.
- Use an object destructuring annotation as the source type for binding properties instead of extracting from a bare initializer like `{}`.
- Add a smoke script covering annotated tuple and optional object destructuring.

## Compatibility impact
- Focused `types/tuple`: +1 (`types/tuple/tupleElementTypes1.ts`).
- Focused `controlFlow`: +1 (`controlFlow/controlFlowDestructuringDeclaration.ts`).
- Full local diff vs `scratch/tsc_next_current_full.txt` (3340/4928): +3 new passes, -2 noisy/pre-existing failures, net +1 in that run.
- Push hook best-observed compliance snapshot: TypeScript 6.0.0 conformance 3342/4928, +1 over committed main docs and +2 over the exact local pre-change dump.

No Test262 change; post-commit and push hooks kept Test262 language/built-ins snapshots unchanged.

## Verification
- `go build ./...`
- `go test ./tests -run TestScripts`
- `./paserati-testtsc -path /Users/nooga/lab/TypeScript -subpath types/tuple -timeout 1s -diff scratch/tsc_next_current_tuple.txt -dump scratch/tsc_next_after_destruct2_tuple.txt`
- `./paserati-testtsc -path /Users/nooga/lab/TypeScript -subpath controlFlow -timeout 1s -diff scratch/tsc_next_current_controlflow.txt -dump scratch/tsc_next_after_destruct2_controlflow.txt`
- `./paserati-testtsc -path /Users/nooga/lab/TypeScript -timeout 1s -diff scratch/tsc_next_current_full.txt -dump scratch/tsc_next_after_destruct2_full.txt`

PR #14, opened 2026-06-28
checker: use destructuring annotations for binding types

nooga:
## Summary

Teach destructuring declarations to use explicit annotations as the binding source type.

- Contextually type annotated array destructuring initializers before checking element bindings.
- Use annotated object destructuring types when extracting optional properties and defaults.
- Add a smoke script covering annotated tuple and object destructuring.
- Refresh generated compliance docs.

## Compat impact

Published TypeScript snapshot: `3341/4928 -> 3342/4928` (`+1`).

Against the exact local pre-change full dump used for this branch, the best observed push-hook result was `3340/4928 -> 3342/4928` (`+2`), with suite-order flakiness noted in the TypeScript runner.

Focused wins:

- `types/tuple/tupleElementTypes1.ts`
- `controlFlow/controlFlowDestructuringDeclaration.ts`

## Validation

- `go build ./...`
- `go build -o paserati-testtsc ./cmd/paserati-testtsc`
- `go test ./tests -run TestScripts`
- Focused `paserati-testtsc` diffs for `types/tuple` and `controlFlow`
- Full `paserati-testtsc` diff against the local pre-change dump

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ go test ./tests/...
 | :-- | --: | --: | --: | --: | --: |
 | Test262 language | 23,141/23,523 | 382 | 0 | 0 | 98.4% |
 | Test262 built-ins | 16,862/23,294 | 6,432 | 0 | 0 | 72.4% |
-| TypeScript 6.0.0 conformance | 3,341/4,928 | 1,114 | 473 | 0 | 67.8% |
+| TypeScript 6.0.0 conformance | 3,342/4,928 | 1,113 | 473 | 0 | 67.8% |
 <!-- compliance:end -->
 
 The Test262 language and built-ins figures come from the local baseline snapshots for the checked-out ECMA-262 conformance tests. The TypeScript figure comes from the single-file conformance runner against the TypeScript 6.0.0 test suite.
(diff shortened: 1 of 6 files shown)


mparrett:
Findings apply to both #13 and #14: they point to the same commit/diff.

- [P1] `pkg/checker/checker.go:3459`: array destructuring still derives binding and rest element types from `valueType`, not the annotation type. Because contextual typing is only applied when the initializer is an `ArrayLiteral`, annotated destructuring from `any` loses the declared tuple types:
  ```ts
  let pair: any = [1, "s"];
  let [a, b]: [number, string] = pair;
  a = "oops"; // accepted on the PR, should be a type error
  ```
  Rest has the same issue:
  ```ts
  let v: any = [1, "s"];
  let [a, ...rest]: [number, string] = v;
  rest = [1]; // accepted, but rest should be string[]
  ```

- [P2] `pkg/checker/checker.go:3615`: object destructuring from tuple annotations now gets a new false positive because `destructureType` may be a `TupleType`, but this object-like check only accepts `ObjectType`, `ArrayType`, or `Any`.
  ```ts
  let v: any = [1];
  let { 0: x }: [number] = v; // PR errors: cannot destructure non-object type '[number]'
  ```

Verification run on the PR branch: `go test ./tests -run TestScripts` passed.

PR #15, opened 2026-06-30
perf page: family-grouped breakdown + cross-tier CPU filter

mparrett:
Improvements to the "are we fast yet?" page from #5. Client-side only — no CI cost, no data changes.

**TL;DR** — just want to see it? https://mparrett.github.io/paserati/

## 1. Family-grouped breakdown

The Breakdown view renders all ~30 benchmarks as one wall of lines (and a 30-item legend), which makes individual movement hard to read. This replaces it with Overall plus per-family tabs — `GetOwn`, `Prototype`, `Workloads` — each showing a bold mean and ±σ band over its faint member lines, with `All` kept as the full breakdown. The legend stays at three entries in every grouped view. Groups are derived from the benchmark key (`pkg.Benchmark<Family>/...`), so there's no hardcoded list.

## 2. Cross-tier CPU filter

The calibration anchor is a pure-ALU loop, but GitHub's hosted runners are heterogeneous — the timeline has landed on EPYC 7763, EPYC 9V74, and Xeon 8370C. Across those microarchitectures the anchor drifts relative to the mixed, memory-bound benchmarks it normalizes: an EPYC 9V74 reads the anchor ~12% slow but not the suite, collapsing that snapshot's ratios into a phantom "dip." This keeps only the points on the modal `cpu_model` and notes how many were hidden in the footer. The underlying snapshots stay on `perf-data`.

## 3. Rebase ratios off the modal snapshot when the baseline is off-tier

The filter above drops off-tier *snapshots*, but the page still divided every series by `baseline.json`'s ratios — and the committed baseline was captured on EPYC 9V74, the exact off-tier the anchor drifts on, so that artifact stayed baked into every plotted value. When `baseline.machine.cpu_model` isn't the modal CPU, the per-benchmark reference now falls back to the first modal snapshot. Overall then centers near 1.0× instead of ~1.15×.

## 4. Overall = the continuous micro suite (Test262 in its own tab)

Overall averaged every non-null series, so the sparse `test262.total` point shifted the mean's composition at exactly the snapshots where it appears. Overall now averages only the continuous micro families (GetOwn + Prototype + Workloads); Test262 keeps its own tab.

## 5. Segment the Test262 series at set-composition changes

`test262.total` is opt-in and sparse, and its passing set drifts between points. The series is now segmented at `set_hash` changes, so dots measured over different sets aren't joined into one continuous line — the timeline-side composition guard, paired with #16 (the metric) and #24 (the `set_hash` source).

## Note

1–2 are the original page work; **3–5 were added in response to the review** of this page and #10 (off-tier normalization, Overall composition, set-comparability). The screenshots below show the family grouping; the later filters and annotations layer on top of it.

## Visuals

image

image

image

image

### DIFF
--- a/docs/perf/index.html
+++ b/docs/perf/index.html
@@ -102,10 +102,7 @@
   </header>
 
   <div class="controls">
-    <div class="toggle" role="group" aria-label="View mode">
-      <button id="m-overall" class="seg" aria-pressed="true">Overall</button>
-      <button id="m-all" class="seg" aria-pressed="false">Breakdown</button>
-    </div>
+    <div id="modes" class="toggle" role="group" aria-label="View mode"></div>
     <span id="count" class="count"></span>
   </div>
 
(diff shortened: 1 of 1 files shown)


mparrett:
Review findings, tested in a browser with all seven real perf-data snapshots:

**[P1] Rebase ratios when the committed baseline is off-tier**

The new filter retains the modal snapshot CPU, but the page still divides every series by baseline.json ratios. In the real data, six snapshots use EPYC 7763 while the committed baseline uses EPYC 9V74, the exact tier this PR describes as having an unreliable anchor. The off-tier artifact therefore remains embedded in every visible value; the browser run showed the Overall series around 1.12 to 1.22 times despite filtering. When baseline.machine.cpu_model differs from modalCpu, please use the first modal snapshot as the reference or provide a same-tier baseline.

**[P1] Exclude sparse Test262 data from Overall**

Overall averages every non-null series at each snapshot. Because PR #16 makes Test262 intentionally sparse, its series joins the mean only at opted-in commits, changing the composition of Overall at those points. With the real snapshots plus a neutral 1.0 Test262 point, the displayed mean moved from 1.1512 to 1.1463 solely because the extra series appeared. Please keep Overall restricted to the continuous micro suite and leave Test262 in its dedicated view.

The page rendered without browser or console errors.

nooga (approved):
## Overview
Single-file change to `docs/perf/index.html`: (1) replaces the hardcoded Overall/Breakdown toggle with dynamically generated per-family tabs (`Overall`, `GetOwn`, `Prototype`, `Workloads`, `Test262`, `All`), each showing faint member lines behind a bold mean ± σ band; (2) filters timeline snapshots to the modal `cpu_model`, hiding cross-tier points that would otherwise distort the anchor-normalized ratios, with a footer note on how many were hidden.

## Correctness
- `groupOf()` splits on `.Benchmark` and matches prefixes (`GetOwn`, `Prototype*`) — verified against the real benchmark key formats produced by `cmd/bench-test262/main.go` / `cmd/bench-ratchet/main.go`. All fall into the intended buckets; unrecognized names fall through to `Workloads`, a reasonable catch-all.
- Modal-CPU filter and `baselineOnTier` guard are logically sound; both handle empty/undefined `machine.cpu_model` and empty-snapshot edge cases without throwing.
- "Overall" mean deliberately excludes the sparse, opt-in `Test262` family so its intermittent presence doesn't shift the mean's composition across snapshots — sound reasoning, well commented.

## Backward compatibility
- One regression: the URL query param changed from lowercase (`?mode=all`/`overall`) to capitalized mode names, and the comparison is case-sensitive. Old bookmarked/shared links using `?mode=all` will silently fall back to `Overall` instead of `All`. Minor, but worth a one-line case normalization if old links are expected to still work.
- No data-format changes; `baseline.json`/`timeline/*.json` are untouched, so this is safe to ship independently of any pipeline changes.

## Style / minor nits
- The CPU-name shortening regex is a narrow heuristic tied to current AMD/Intel naming; harmless no-op if it doesn't match a future CPU string.
- Toggle group now renders 6 buttons instead of 2 — no new responsive/overflow handling was added for narrow viewports; likely acceptable but worth a glance on mobile.

## Test coverage
No automated tests — reasonable given this is a static, client-rendered dashboard page with no build step or CI hook.

mparrett:
Sorry — this dismissed your approval. After you approved (at `c41e7ab`) I pushed one more commit (`52657a1`) folding in the two items from your review:

- **`?mode=` regression:** the tab rename made the query param case-sensitive, so old `?mode=all` / `?mode=overall` links fell back to Overall. It now matches case-insensitively, so those links resolve again.
- **`set_hash` segmentation (#23 follow-on):** on the Test262 tab, where a point's `set_hash` differs from its predecessor the mean isn't over the same passing set — so the line now dashes that span and marks the point, instead of drawing it as one continuous series.

Browser-verified both (dashed span lands on the set-change boundary, mode links resolve, no console errors). Nothing else changed from the approved diff. Could you re-approve when you get a chance? Thanks!

nooga (approved):
Thanks Matt — the family-grouped tabs and CPU-tier filtering are a nice readability win, and it degrades gracefully when set_hash isn't present yet. Merging.

PR #17, opened 2026-07-07
compiler: report oversized-function jumps as a compile error, not a panic

mparrett:
## Problem

`patchJump` / `patchJumpToTarget` panic when a jump offset exceeds the int16
branch encoding (±32767 bytes):

```go
if offset > math.MaxInt16 || offset < math.MinInt16 {
    // TODO: Add proper error handling instead of panic
    panic(fmt.Sprintf("Compiler error: jump offset %d exceeds 16-bit limit", offset))
}
```

This is a hard crash with a Go stack trace on **valid input** — any single
function large enough that a branch spans more than 32 KB of bytecode. A big
`switch`, a long `if`/`else` chain, or machine-generated / bundled code all reach
it. Minimal repro:

```js
function big(x) {
  let s = 0;
  if (x > 0) {
    s = s + 0; s = s + 1; /* …~12k statements… */
  }
  return s;
}
big(1);
```

```
panic: Compiler error: jump offset 132005 exceeds 16-bit limit
        …/pkg/compiler/compiler.go:2991 patchJump
```

The code already flagged this with a `TODO: Add proper error handling instead of
panic`.

## Fix

Route both overflow sites through a new `recordJumpOverflow`, which appends a
positioned `CompileError` and returns instead of panicking:

```
PS3001 [ERROR]: function too large: a jump offset of 132005 bytes exceeds the
16-bit branch limit (±32767); split this function into smaller ones
  3:   if (x > 0) {
     ^
    at line 3, column 0
```

The unpatched placeholder left behind is harmless: a non-empty error list aborts
the compile before the chunk can run, so no malformed bytecode executes. No
signatures change (the 126 `patchJump` call sites are untouched) — the error goes
through the compiler's existing `c.errors` accumulator.

## Testing

- New `TestJumpOffsetOverflowIsGracefulError` compiles a ~6000-statement function
  and asserts a graceful "function too large" compile error — and, by not
  panicking, that the crash is gone.
- `go test ./pkg/compiler/` and the `TestScripts` smoke suite stay green; `go vet`
  clean.

## Scope / follow-up

This makes oversized functions fail gracefully; it does not make them *compile*.
Actually compiling them would need 32-bit long-jump opcodes plus branch
relaxation in the emitter — a larger, separate change. This PR is the minimal
robustness fix: don't crash the process on valid input.

### DIFF
--- a/pkg/compiler/compiler.go
+++ b/pkg/compiler/compiler.go
@@ -2986,14 +2986,25 @@ func (c *Compiler) patchJump(placeholderPos int) {
 	}
 
 	if offset > math.MaxInt16 || offset < math.MinInt16 { // Use math constants
-		// Handle error: jump offset too large
-		// TODO: Add proper error handling instead of panic
-		panic(fmt.Sprintf("Compiler error: jump offset %d exceeds 16-bit limit", offset))
+		c.recordJumpOverflow(placeholderPos, offset)
+		return
 	}
 
 	// Write the 16-bit offset back into the placeholder bytes (Big Endian)
 	c.chunk.Code[operandStartPos] = byte(int16(offset) >> 8)     // High byte
 	c.chunk.Code[operandStartPos+1] = byte(int16(offset) & 0xFF) // Low byte
+}
+
+// recordJumpOverflow reports a jump whose distance exceeds the 16-bit branch
+// offset — i.e. a function too large for the bytecode format — as a graceful
+// compile error instead of panicking. The unpatched placeholder is harmless: a
+// non-empty error list aborts the compile before the chunk can run.
+func (c *Compiler) recordJumpOverflow(placeholderPos, offset int) {
+	c.errors = append(c.errors, &errors.CompileError{
+		Position: errors.Position{Line: c.chunk.GetLine(placeholderPos)},
+		Msg: fmt.Sprintf("function too large: a jump offset of %d bytes exceeds the 16-bit "+
+			"branch limit (±32767); split this function into smaller ones", offset),
+	})
 }
 
 // patchJumpToTarget patches a jump to a specific target PC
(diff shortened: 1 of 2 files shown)


nooga (approved):
## Overview
Converts the `patchJump`/`patchJumpToTarget` panic (`"jump offset %d exceeds 16-bit limit"`) into a graceful `errors.CompileError` via a new `recordJumpOverflow` helper. I built the branch, ran the new test with and without the fix, and ran the CLI against a synthetic 6000-statement function to confirm end-to-end behavior.

## Verification performed
- Confirmed the overflow bounds check (`> MaxInt16` / `< MinInt16`) is correct and unchanged from before.
- Both call sites now `return` immediately after recording the error, before writing to the placeholder bytes — no malformed bytecode is emitted.
- Reverted just `compiler.go` in an isolated worktree and reran `TestJumpOffsetOverflowIsGracefulError`: it panics on the old code and passes cleanly on the new code, so the test is a real regression guard.
- Built the CLI and ran it against a generated 6000-statement function: got a clean `PS3001 [ERROR]: function too large...` message with exit code 70, no Go stack trace.
- `go test ./tests -run TestScripts` passes (smoke suite unaffected).

## Minor nitpicks (non-blocking)
- `recordJumpOverflow` only populates `Position.Line`, leaving `Column`/`StartPos`/`EndPos` at zero, so the printed caret points at column 0 rather than the actual branch site — a real limitation of `Chunk.GetLine` (no column tracking), not something introduced carelessly.
- For very large functions, multiple jumps could overflow in sequence, potentially producing several redundant "function too large" errors in one compile. Not incorrect, just possible message noise.

## Scope
The PR explicitly does not attempt to make oversized functions compile (that would require 32-bit long jumps / branch relaxation, similar in spirit to `OpClosure16`). That's a reasonable, clearly-documented scope boundary for a "stop crashing on valid input" fix.

nooga (approved):
Thanks Matt — clean fix, and the regression test compiling a ~6000-statement function to exercise the overflow path is exactly what this needed. Merging.

PR #18, opened 2026-07-09
compiler: anonymous arrows capturing >255 variables emit OpClosure16

mparrett:
## What

`compileArrowFunctionLiteral` — the path that compiles arrows in callback or
IIFE position — emits `OpClosure` with an 8-bit upvalue count. Past 255
captured variables the count truncates (300 reads back as 44), and the VM
decodes the remaining upvalue descriptors as opcodes, corrupting the chunk.
The named-declaration path already guards this in `emitClosure` with
`OpClosure16`; this brings the anonymous-arrow path in line, including the
`CaptureFromUpvalue16` 16-bit descriptor form for enclosing upvalue indexes
past 255. No VM change — the decoder already handles both encodings.

## How it shows up

Found while stress-testing large generated closures. Hand-written code rarely
captures 256 variables in one arrow, but generated or bundled callbacks can.
Pre-fix the failure mode depends on what the mis-decoded descriptor bytes
happen to encode: invalid-upvalue-index errors in the good case, silently
wrong opcodes in the bad one.

## Test

`tests/scripts/closure_arrow_upvalues_256.ts`: a callback-position arrow
capturing 300 function-local variables, split across two nesting levels so no
function's locals spill and the count truncation is the only variable. Fails
before the fix, passes after. `go test ./tests ./pkg/compiler ./pkg/vm` is
green.


nooga (approved):
## Overview
Solid, well-targeted fix. `compileArrowFunctionLiteral` had its own duplicate `OpClosure` emission (separate from `emitClosure`/`emitClosureGeneric`) that never widened for >255 upvalues. This PR adds the `OpClosure16` header and `CaptureFromUpvalue16` descriptor form, matching the existing scheme exactly (confirmed byte order and VM decode logic already support both — no VM changes needed).

## Verification performed
- Built the PR branch in an isolated worktree, ran `go test ./tests -run TestScripts`, `./pkg/compiler`, `./pkg/vm` — all green.
- Disassembled `tests/scripts/closure_arrow_upvalues_256.ts`: confirmed `OpClosure16 ... Upvalues 300` is emitted for the arrow, correctly widened.
- Confirmed the "anonymous arrows" scoping is accurate: `compileArrowFunctionWithName` (named arrows) already routes through `emitClosureGeneric`, which had this fix; `compileArrowFunctionLiteral` was the only remaining unpatched path.

## Findings
1. **Test coverage gap (non-blocking):** the new test exercises the `OpClosure16` header/count-truncation fix but not the second hunk in this diff — the `enclosingFreeIndex > 255` → `CaptureFromUpvalue16` branch. Disassembly confirms all 300 descriptors in the test use indices ≤ 150. Exercising that branch needs 3+-level nesting where an intermediate function's own upvalue list exceeds 255 entries. Given it mirrors already-tested `emitClosure` code and shares the VM decoder, risk is low — but consider a follow-up test.
2. **Pre-existing disassembler gap (out of scope, worth a ticket):** `closureInstruction` in `pkg/vm/bytecode.go` has no explicit `case` for `captureType == 4` (`CaptureFromUpvalue16`) — falls to `default`, mislabeling and under-consuming bytes in `-bytecode` output (execution unaffected). This PR makes that path more reachable.
3. **Minor duplication:** `compileArrowFunctionLiteral` keeps its own third copy of the upvalue-descriptor-emission logic instead of delegating to `emitClosureGeneric`. Not a blocker for a bug-fix PR, but a candidate for later consolidation.

No regressions to the existing 8-bit path; boundary at 255/256 is correct.

nooga (approved):
Thanks Matt — traced every OpClosure emission site and confirmed this closes the one gap (arrow literals) that the existing 16-bit encode/decode path didn't already cover; root cause fully fixed, not just the repro. Found one pre-existing, unrelated bug while in there: the -bytecode disassembler is missing a case for CaptureFromUpvalue16, corrupting debug output only (not execution) — will file a follow-up separately, not blocking this. Merging.

PR #19, opened 2026-07-10
compiler: fix spilled let/const bindings losing their value

mparrett:
## Summary

A binding that spills (declared past the 256-entry register file) never has its real value written into its spill slot, so a later read returns the TDZ sentinel (`TypeUninitialized`, which prints as "type 33"). Symptoms: `TypeError: <unknown type: 33> is not a function` when a spilled closure is called, or `ReferenceError: Cannot access variable before initialization` — same defect.

## Root cause

The block-predefine pass seeds a spilled `let`/`const` with a TDZ marker in its slot. Several initializer paths then write the real value to the wrong place: they reuse the spilled symbol's `Register` field, which is `0` for a spilled symbol, and since `nilRegister` is `255` the "already has a register" check (`Register != nilRegister`) passes. The value lands in `R0` and the slot keeps the sentinel.

## Fix

Honor the predefined spill slot on **every** initializer path that can receive a spilled binding — store the value (or `undefined`) into the slot instead of `R0`. The invariant is now explicit: once the predefine pass spills a binding, every write targets that slot and the symbol stays spilled. Covered:

- **let/const closures** (`let f = () => …`, plus `function`/`const` variants) — via two small store-to-slot helpers.
- **var closures** (`var f = () => …` / `var f = function(){}`).
- **class expressions** (`let`/`const`/`var C = class {}`) — compiled into a temp, written back to the slot.
- **uninitialized `let x;`** — stores `undefined` into the slot instead of leaving the sentinel.

## Extended after @nooga's review

The first-round fix covered only let/const closures. Review flagged that `var f = () => …` still crashed, and my own pre-review pass found the class-expression and no-initializer cases — same root, different init paths. This version folds in that whole family. The audit also surfaced a **distinct** pre-existing bug (fused `++`/`--` on a spilled local targets R0, `compile_expression.go`) — that's a *use* path, not a binding-init path, so it's out of scope here and filed as #27.

## Tests

Consolidated the two ~500-line single-scenario spill tests into one documented matrix (`spill_binding_matrix_test.ts`) exercising the whole family past the spill threshold: var/let/const closures, class expressions, an uninitialized `let`, a reassigned closure, and a closure capturing spilled locals. Verified fail-before / pass-after on every form; `TestScripts` green; `go build`/`vet`/`gofmt` clean.

The first commit is a standalone gofmt of the touched file (comment alignment); the fixes follow.


mparrett:
I found two remaining spill-path bugs while reviewing this.

1. Spilled class-expression bindings are still left uninitialized.

`let C = class {}` and `const C = class {}` still go through the class-specific branches in `compile_statement.go`, which allocate/compile into a temp register when the binding was already predeclared as spilled, but never store that value back to `sym.SpillIndex`. The later predefined-symbol path also checks `sym.Register != nilRegister` without excluding `sym.IsSpilled`, so the class value gets moved into `R0` instead of the spill slot.

Relevant areas:
- `pkg/compiler/compile_statement.go` `let` class branch around lines 125-148
- `pkg/compiler/compile_statement.go` `let` final define path around lines 246-248
- `pkg/compiler/compile_statement.go` `const` class branch around lines 696-717
- `pkg/compiler/compile_statement.go` `const` final define path around lines 764-765

Repro against the PR build:

```ts
function outer() {
  let v0 = 0; /* ... v1 through v220 ... */
  let C = class {};
  return [typeof C, C === undefined].join(",");
}
outer()
```

Actual: `object,false`
Expected: `function,false`

Same behavior reproduces with `const C = class {}`.

2. Spilled `let x;` declarations without initializers also miss the spill slot.

The implicit-`undefined` path checks `sym.Register != nilRegister` without excluding spilled symbols. For a predeclared spilled `let`, that writes `undefined` into `R0` and leaves the real spill slot holding the TDZ sentinel/object value. `InitializeTDZ` then clears the compile-time TDZ flag, so later reads use the stale spill slot.

Relevant area:
- `pkg/compiler/compile_statement.go` implicit undefined path around lines 196-228, especially 212-214

Repro:

```ts
function outer() {
  let v0 = 0; /* ... v1 through v220 ... */
  let x;
  return [typeof x, x === undefined].join(",");
}
outer()
```

Actual: `object,false`
Expected: `undefined,true`

I also checked the main closure paths this PR is targeting. Function/arrow closure spill cases, including recursive self-reference, returned the expected value in my local PR build.

nooga (changes requested):
## Overview
Solid, well-explained fix for a genuine correctness bug: spilled `let`/`const` closure bindings (and captured spilled locals in general) were losing their initial value because two initializer paths reused a spilled symbol's zero-valued `Register` field instead of honoring `IsSpilled`/`SpillIndex`. I built the branch, ran the smoke suite (green), and reproduced both symptoms on `main` (`type 33 is not a function` for the closure-register case, and the plain-captured-variable case) then confirmed they pass on this branch.

## What's good
- Root cause is correctly identified and the fix addresses it directly (not a workaround).
- `closureBindingDest`/`finalizeClosureBinding` helpers are clean, reused across all 4 let/const × function/arrow call sites, and match the existing "store-to-spill-slot" idiom used elsewhere in the compiler (`compile_assignment.go`, hoisted-function handling in `compiler.go`).
- The const-specific branch correctly uses `InitializeTDZ` (preserving `IsConst`) rather than blindly calling `DefineSpilled` — shows real attention to let vs. const semantics.
- New tests are genuine regressions (verified fail-before/pass-after), and multi-closure-over-same-spilled-var / post-creation-reassignment scenarios work correctly (verified manually).

## Blocking issue: the fix is incomplete
The same bug class still reproduces for `var` declarations initialized with a function/arrow literal, because `compileVarStatement`'s closure branches have the identical missing `!sym.IsSpilled` guard and were not touched by this PR:

```ts
function outer(): number {
  let v0 = 0; /* ...300 lets... */ let v299 = 299;
  var f = (): number => v0 + v1 + v2;
  return f();   // TypeError: undefined is not a function — still broken on this branch
}
```

Both the arrow and plain-function-literal variants crash. Since this PR's stated goal is to close out this class of defect, please either fold in the equivalent fix for `compileVarStatement`'s closure paths, or explicitly split it into a tracked, named follow-up issue referenced in this PR's description so it isn't lost.

## Minor / non-blocking
- The two new test scripts are large blocks of mechanically-repeated `let vN = N;` lines; consider a shorter reproduction (or a comment documenting the minimum spill threshold) for maintainability.

mparrett:
Addressed in `4e63774`.

Extended the fix to the three remaining paths in the same family, each now storing into the spill slot via the same `IsSpilled`-first pattern the closure/const-value paths already used:
- `var f = () => …` / `var f = function(){}` — your blocking repro
- `let`/`const C = class {}` — compiled into a temp, never stored back
- `let x;` uninitialized — wrote `undefined` to R0, leaving the slot's sentinel live

Also guarded the var-no-init hoist-skip against a spilled symbol so it routes to the `IsSpilled` branch instead of mistaking R0 for a real register.

Verified fail-before/pass-after on all four forms plus a reassigned closure and a closure capturing spilled locals. `TestScripts` green; build/vet/gofmt clean.

**Test nit:** consolidated the two ~500-line single-scenario files into one documented matrix (`spill_binding_matrix_test.ts`, −835/+163) covering the whole family.

**Out of scope:** the audit surfaced a distinct pre-existing bug — fused `++`/`--` on a spilled local targets R0 (`compile_expression.go:705,719`), filed as #27. Not a binding-init path, so left out here.

nooga (approved):
Thanks Matt — good catch and a real root-cause fix: the spilled symbol's Register defaulting to 0 (Go zero value) was indistinguishable from a legitimate R0 allocation, and the IsSpilled check now guards every initializer path (let/const/var closures, class expressions, uninitialized let). Verified empirically pre/post-fix. One suspected (unconfirmed) related gap from code inspection: closures over a *spilled* per-iteration for-loop let may share one slot instead of getting a fresh binding per iteration, per the comment at compile_statement.go:1068 vs CaptureFromSpill's capturability — worth a tracked follow-up, not blocking this fix. Merging.

PR #20, opened 2026-07-12
perf(bench): VM hot-path benchmark harness

mparrett:
Benchmark scaffolding only — no VM changes. This is the base of a short stack of hot-path optimizations; landing the harness first means each follow-up is a pure `pkg/vm` diff with a before/after on benchmarks that already exist on `main`, instead of every opt PR re-touching the shared bench registry.

## What's here
- `tests/bench_test.go`: `BenchmarkAdd` / `BenchmarkArith` / `BenchmarkSetIndex`, driving small `bench_*.ts` scripts through the public driver API.
- `pkg/vm/isobject_bench_test.go`: `BenchmarkIsObject` over a hot-path value mix, plus `TestIsObjectExhaustive` locking the `IsObject == [TypeObject, TypeProxy]` range invariant (passes on the current OR-chain).
- `pkg/vm/tointeger_bench_test.go`: `BenchmarkToInteger`.

## Baseline (this commit, one full script run)
`BenchmarkSetIndex`: 6,596,688 B/op, 777,935 allocs/op — the target the OpSetIndex allocation-elimination change (next in the stack) drives down.

## Follow-ups (staged, will open once this lands)
1. OpSetIndex per-write allocation elimination
2. IsObject OR-chain → range check
3. Number fast paths (OpAdd / arithmetic / ToInteger)

`TestScripts` and `go test ./pkg/vm` green.

### DIFF
--- a/pkg/vm/isobject_bench_test.go
+++ b/pkg/vm/isobject_bench_test.go
@@ -1,0 +1,62 @@
+package vm
+
+import "testing"
+
+// isObjectMix models the hot-path distribution: IsObject is called predominantly
+// on non-object values (numbers in arithmetic/ToInteger guards), which is the
+// OR-chain's worst case (all comparisons fail before returning false).
+func isObjectMix() []Value {
+	vs := make([]Value, 0, 32)
+	for i := 0; i < 24; i++ {
+		vs = append(vs, IntegerValue(int32(i))) // 24 numbers (typ far below the object range)
+	}
+	vs = append(vs,
+		NewString("x"), BooleanValue(true), Undefined, Null, // more non-objects
+		Value{typ: TypeObject}, Value{typ: TypeArray},
+		Value{typ: TypeProxy}, Value{typ: TypeMap}, // a few objects
+	)
+	return vs
+}
+
+func BenchmarkIsObject(b *testing.B) {
+	vs := isObjectMix()
+	b.ResetTimer()
+	var sink int
+	// Inner range loop so IsObject dominates ns/op instead of index arithmetic.
+	// ns/op therefore covers len(vs) IsObject calls; consistent across A/B runs.
+	for i := 0; i < b.N; i++ {
+		for j := range vs {
+			if vs[j].IsObject() {
+				sink++
+			}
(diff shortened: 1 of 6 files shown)


nooga (approved):
### Overview
Solid, low-risk scaffolding PR — pure additions, no production code touched. Verified `go build ./...`, `go vet ./...`, `TestScripts`, and the two new correctness-locking tests (`TestIsObjectExhaustive`, `TestToIntegerEquivalence`) all pass. Manually ran all five new benchmarks; they execute correctly and roughly reproduce the alloc numbers quoted in the description.

### Strengths
- `TestIsObjectExhaustive` independently verified correct: `[TypeObject, TypeProxy]` is exactly contiguous with the 18 types in the current OR-chain — a genuinely useful guard for the planned range-check refactor.
- Timer placement (`b.ResetTimer()` after `compileFile`) matches existing conventions; no setup leaking into timed sections.
- Auto-discovery in `cmd/bench-ratchet` already covers `pkg/vm` and `./tests`, so these benchmarks integrate with the ratchet baseline for free.

### Nits (non-blocking)
1. `tests/bench_test.go`: the old `// BenchmarkMatrixMult runs the matrix_mult.ts script.` comment is now orphaned above `BenchmarkSetIndex` — please move it back above `func BenchmarkMatrixMult`.
2. `tests/scripts/bench_arith.ts` asserts `// expect: 0` on `s - s`, which is always zero regardless of whether the arithmetic is correct. Consider a real checksum (as `bench_setindex.ts` does) so this fixture actually guards the upcoming arithmetic fast-path change.

Approving as-is; the nits can be folded into this PR or an immediate follow-up.

nooga (approved):
Thanks Matt — solid scaffolding with invariant-locking tests (TestIsObjectExhaustive, TestToIntegerEquivalence) alongside the benchmarks. One tiny non-blocking nit: the new BenchmarkSetIndex landed between the // BenchmarkMatrixMult doc comment and that function, so the comment now describes the wrong benchmark — worth a quick follow-up whenever convenient. Merging as-is.

Issue #23, opened 2026-07-12
Epic: perf benchmarking infrastructure (tracker)


mparrett:
Master tracker for the "are we fast yet?" benchmarking stack — the same-runner A/B PR gate, the Test262 macro-benchmark, and the trend page. Ties together the merged foundation, the in-flight PRs, and the review findings that gate them.

## PRs

**Merged foundation**
- #4 — bench-ratchet + same-runner A/B PR check (anchor-normalized micro gate)
- #5 — benchmark trend page ("are we fast yet?")
- #10 — Test262 macro-benchmark (opt-in same-runner A/B, `perf-test262` label)

**In flight**
- #15 — perf page: family-grouped breakdown + cross-tier CPU filter (F1+F2 fixes local, unpushed)
- #16 — perf-timeline: opt-in `test262.total` macro point (wires #10's tooling into #15's page)
- #20 — VM hot-path benchmark harness
- #22 — bench-ratchet: reduce `-count` samples by min, not mean
- #24 — bench-test262 set fingerprint (F3)
- #25 — perf-test262 corpus pin + drop paths filter (F4+F5)

**Open finding**
- #21 — perf-pr A/B is ungateable as a single shot; median-of-N fixes it

## Dependency shape

```
#4 (A/B gate) ── #10 (test262 A/B) ── #16 (test262.total on the page)
     │                                    │
#5 (page) ─────── #15 (page grouping + CPU filter) ──┘
```

#16 is the integration point: it makes Test262 a **sparse, opt-in** series on the page. Three of the findings below are latent until #16 lands, which is why they're worth clearing while the series contract is still open.

## Findings to resolve

From the retrospective review of #10 and the review of #15 (both 2026-07-12).

- [x] **F1 · #15 · P1 — off-tier baseline.** Ratios still divide by `baseline.json`, whose `cpu_model` is EPYC 9V74 (the exact off-tier the PR calls unreliable) while snapshots are EPYC 7763. The filter drops off-tier *snapshots* but the artifact stays embedded in every value (Overall ~1.12–1.22× in-browser). → When `baseline.machine.cpu_model !== modalCpu`, rebase onto the first modal snapshot (or ship a same-tier baseline). **Fixed** in `perf/page-grouping` (commit `983d9c3`): verified against 22 real snapshots, Overall 1.13–1.21× → 0.93–1.00×.
[...]

PR #24, opened 2026-07-13
bench-test262: fingerprint the contributing test set so test262.total is comparable

mparrett:
Resolves finding **F3** from the retrospective review of #10 (tracked in #23).

## The problem

The Test262 macro emits `test262.total` as **sum + count** of the passing, non-timed-out tests. That loses test identity: two captures with equal counts can cover *different* sets — a test flipping pass→timeout while another flips back keeps the count but changes both the set and the sum. The A/B comparator sidesteps this by diffing the per-test merged files, but the aggregate total advertised for the timeline (#16) can't, so a composition change reads as an engine change.

## The fix

Each `StreamRecord` now carries a **`set_hash`**: an order-independent, truncated SHA-256 over the exact set of test paths it summed.

- **Producer** (`cmd/bench-test262`): computes the hash per emitted record.
- **Schema** (`pkg/perfdata`): `SetHash` on `StreamRecord` and `BenchmarkEntry`, `omitempty` so ordinary micro benchmarks stay unaffected.
- **Aggregate** (`cmd/bench-ratchet`): threads the hash onto the snapshot's `BenchmarkEntry`. Disagreeing hashes across a key's `-count` samples blank out — no coherent identity — which is the honest signal for a flaky set.

A consumer can then gate: a `test262.total` point whose `set_hash` differs from the reference isn't comparable.

## Verification

- Unit tests (`cmd/bench-test262/main_test.go`): equal-count/different-set → different hash; order-independent; failed/timed-out excluded from both sum and set.
- End-to-end: a `set_hash` on a `test262.total` record survives aggregate into the snapshot entry alongside `ratio_to_anchor`, and correctly stays off per-sample rows.
- `go build ./...`, `go vet ./...` clean.

## Follow-on

**#16 should gate its `test262.total` series on `set_hash`** — this PR gives it the tool; the page-side drop/flag is #16's job.

### DIFF
--- a/cmd/bench-ratchet/main.go
+++ b/cmd/bench-ratchet/main.go
@@ -110,6 +110,7 @@ type Result struct {
 	NSPerOp     float64
 	BytesPerOp  int64
 	AllocsPerOp int64
+	SetHash     string // identity of the aggregated set (test262 macro); empty for micro benchmarks
 	Samples     []BenchmarkSample
 }
 
@@ -654,6 +655,8 @@ func aggregateFromFile(path string) (Baseline, error) {
 		count                     int
 		nsSum, bytesSum, allocSum float64
 		iters                     int64
+		setHash                   string // shared set identity across this key's records
+		setHashConflict           bool   // records disagreed → no coherent identity
 		samples                   []BenchmarkSample
 	}
 	byName := map[string]*accum{}
@@ -676,6 +679,13 @@ func aggregateFromFile(path string) (Baseline, error) {
 		if rec.Iterations > a.iters {
 			a.iters = rec.Iterations
 		}
+		if rec.SetHash != "" && !a.setHashConflict {
+			if a.setHash == "" {
+				a.setHash = rec.SetHash
+			} else if a.setHash != rec.SetHash {
+				a.setHash, a.setHashConflict = "", true
+			}
+		}
 		a.samples = append(a.samples, rec.Sample())
 	}
 
@@ -688,6 +698,7 @@ func aggregateFromFile(path string) (Baseline, error) {
 			NSPerOp:     a.nsSum / float64(a.count),
 			BytesPerOp:  int64(a.bytesSum / float64(a.count)),
 			AllocsPerOp: int64(a.allocSum / float64(a.count)),
+			SetHash:     a.setHash,
 			Samples:     append([]BenchmarkSample(nil), a.samples...),
 		})
 	}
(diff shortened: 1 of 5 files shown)


nooga (approved):
**Overview**: Adds `SetHash` fingerprinting to `cmd/bench-test262` and threads it through `cmd/bench-ratchet`'s aggregation into `pkg/perfdata`'s schema, so downstream consumers can distinguish "same count, different test set" from a real change.

**Analysis**: Hash construction (sorted paths, newline-delimited, truncated SHA-256) is correct and collision-safe for this use case. `-filter`/`-subpath` are handled correctly since the hash operates on the already-filtered result set. The aggregator's conflict-blanking logic (differing hashes across `-count` samples → blank, an honest "no coherent identity" signal) is sound, though untested directly.

**Suggestions**: Consider adding a focused unit test in `cmd/bench-ratchet` for the hash-conflict-blanking branch, since it's the one new piece of control flow without direct coverage — not blocking.

**Risks**: None material; change is additive, `omitempty`-guarded, doesn't require a schema version bump, and doesn't touch the interpreter/checker/compiler/VM. `go build ./...` and `go vet ./...` are clean, and `cmd/bench-test262/main_test.go`'s new tests all pass.

mparrett:
**Heads-up: composes with #22, one small merge conflict.** Both this PR and #22 edit `cmd/bench-ratchet/main.go`'s `aggregateFromFile`. Verified by composing all open PRs on a clean nooga/main base — this is the *only* conflict across the set, and it's a 3-line resolution in the `Result{...}` build: take #22's reducer form **and** keep this PR's `SetHash`:

```go
NSPerOp:     reduce(a.nsMin, a.nsSum, a.count),
BytesPerOp:  int64(reduce(a.bytesMin, a.bytesSum, a.count)),
AllocsPerOp: int64(reduce(a.allocMin, a.allocSum, a.count)),
SetHash:     a.setHash,
```

The accumulator auto-merges (each PR adds distinct fields); only this block collides. Either order is fine — whichever lands second applies the above. (This PR should also precede #16, which consumes `set_hash`.)

mparrett:
Folded in your suggestion (`fa8ccf1`): a focused `cmd/bench-ratchet` test for the set_hash conflict-blanking branch — disagreeing `-count` repetitions blank the aggregated identity to `""`, agreeing ones preserve it. Only touches the new test file, so the #22 conflict region is unchanged.

nooga (approved):
Thanks Matt — solid, well-tested fix (order-independent fingerprint, clean pass/fail/timeout exclusion coverage). Merging.

PR #29, opened 2026-07-22
perf(vm): iterator fast path — step built-in iterators without a call or result object

mparrett:
`for-of` over the built-in iterables spent a method call and a `{value, done}` allocation per element. When the loop's cached `next` is a recognized built-in iterator, step it directly with no call and no result object.

## What this does

- **Two opcodes** (`OpIterFastCheck` 172, `OpFastIterNext` 173). The compiler already caches the iterator's `next` in a register once per loop (matching the spec's `IteratorRecord.[[NextMethod]]` caching). `OpIterFastCheck` inspects that cached value once; `OpFastIterNext` advances the built-in iterator state in place. Anything unrecognized — a user iterator, a replaced `next` — fails the check and takes the unchanged generic call path.
- **Coverage** grows across the three commits: plain arrays first, then string / arguments / array-likes / `keys()` / `entries()` (via a shared `BuiltinIterState` with a per-kind `Step`), then Map / Set iterators.
- **Design point for review:** Map/Set iterators share a single `next` on their prototype, so per-iterator state can't ride the `next` closure the way the array family's does. It lives in a new `PlainObject.internalIterState` pointer (+8 bytes per object). Offsetting correctness win: those slots were previously *enumerable user-visible properties*; moving them into the struct makes them unobservable internal slots (spec-correct) and the brand check O(1).

`Step` re-reads the source length/content every iteration, so growth, truncation, and hole normalization during iteration behave exactly like the native `next` it bypasses (both call the same `Step`).

## Numbers

Local (M2): for-of over a plain array 4.1×, string iteration 2.4×, Set iteration 2.6×, Map for-of ~10% (destructuring-bound; see the destructuring PR). The `perf`-label same-runner A/B is the authoritative check.

## Verification

`TestScripts` green; `tests/scripts/for_of_array_fast_path.ts` and the builtin/Map-Set iterator scripts cover hole normalization, length re-read on growth/truncation, shared state with manual `next()`, deopt on replaced `next`, and reuse after early break. Test262 language 0 new failures (differential vs upstream/main control); targeted built-ins 0 new failures.

---

### DIFF
--- a/pkg/builtins/array_init.go
+++ b/pkg/builtins/array_init.go
@@ -2992,28 +2992,33 @@ func createArrayIterator(vmInstance *vm.VM, array *vm.ArrayObject) vm.Value {
 	iterator := vm.NewObject(vmInstance.ArrayIteratorPrototype).AsPlainObject()
 	iteratorVal := vm.NewValueFromPlainObject(iterator)
 
-	// Iterator state: current index
-	currentIndex := 0
+	// Iterator state, shared between the next closure below and the VM's
+	// for-of fast path (OpArrayIterNext reaches it through the next method's
+	// NativeFunctionObject), so manual next() calls and a for-of over the
+	// same iterator advance one index.
+	state := &vm.ArrayIterState{Arr: array}
 
 	// Add next() method to iterator
-	iterator.SetOwnNonEnumerable("next", vm.NewNativeFunction(0, false, "next", func(args []vm.Value) (vm.Value, error) {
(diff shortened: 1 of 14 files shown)


nooga (approved):
Thanks Matt — the fast-path gating is fail-safe (only triggers on Paserati's own IterState-tagged natives, anything else falls to the untouched general path), and .return()-on-break/throw is preserved since cleanup keys off the iterator object itself, independent of the fast path. Verified this manually too. Merging.

PR #30, opened 2026-07-22
perf(vm): store primitive strings inline in Value

mparrett:
Primitive strings were boxed in a heap-allocated `StringObject{value string}` (its embedded `Object` is empty, so the box was just a 16-byte string header on the heap), with `Value.payload` unused.

## What this does

Store the string header directly: `obj` holds the backing pointer (via `unsafe.StringData`), `payload` holds the length. `AsString` reconstructs with `unsafe.String` — no copy, no indirection — and `NewString` no longer allocates. The empty string stores a nil pointer (`payload 0`). The GC scans the `unsafe.Pointer` field, so the backing stays live while the Value is reachable; retention matches the old box. `StringObject` is removed (the `new String()` wrapper is a separate `PlainObject` path, untouched). All primitive-string access already funnels through `NewString` / `AsString` / `ToString`, so the change is three sites in one file.

**Design point for review:** the `unsafe.StringData`/`unsafe.String` round-trip and the GC-liveness argument (the GC scans the `unsafe.Pointer` field, so the backing outlives the Value).

## Numbers

Local (M2): ~23% fewer GC cycles on a concat loop (13 → 10), ~7% wall-clock on string-creation-heavy code; strings built from an existing backing (substring, key materialization) now allocate nothing. The `perf`-label A/B is authoritative.

## Verification

`TestScripts` + unit tests green; vm tests under `-race` (checkptr) clean; a GOGC=1 20k-string stress kept 20000/20000 intact after aggressive GC; empty-string edge cases correct. Test262 language 0 new failures (differential vs upstream/main control); targeted string built-ins 0 new failures.

---

### DIFF
--- a/pkg/vm/value.go
+++ b/pkg/vm/value.go
@@ -151,11 +151,6 @@ func (vt ValueType) String() string {
 	}
 }
 
-type StringObject struct {
-	Object
-	value string
-}
-
 type SymbolObject struct {
 	Object
 	value          string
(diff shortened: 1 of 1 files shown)


nooga (approved):
Thanks Matt — the inline string encoding (unsafe.StringData/unsafe.String, full uint64 payload so no capacity overflow risk, correct GC-liveness reasoning via the obj field) is well thought through. Merging.

PR #31, opened 2026-07-22
refactor(vm): fold the heap initialized[] flag into a value sentinel

mparrett:
Perf-neutral cleanup — framed as simplification + memory, not speed.

## What this does

Folds the heap's parallel `initialized []bool` into a value sentinel so `Heap.Get` reads one array instead of two; a payload-tagged empty marker distinguishes never-set from TDZ. The intended cache-line saving didn't materialize (hot globals stay cached; A/B showed no wall-clock change), but it removes a slice per heap/realm and one storage mechanism. Self-contained in `pkg/vm/heap.go` plus two call-site swaps.

**Design point for review:** the payload-tagged empty sentinel distinguishing never-set from TDZ.

## Numbers

No wall-clock change (A/B neutral). Removes one `[]bool` per heap/realm.

## Verification

Byte-identical behavior; `TestScripts` green; Test262 language 0 new failures (differential vs upstream/main control), 0 regressions.

Lowest priority of the series — take it only if the cleanup is welcome, or hold until a perf item with a real number gives it cover.

---

### DIFF
--- a/pkg/vm/heap.go
+++ b/pkg/vm/heap.go
@@ -4,20 +4,37 @@ import (
 	"fmt"
 )
 
+// heapEmptySlot marks a global slot that has never been explicitly Set (or was
+// deleted). It shares TypeUninitialized with the public TDZ marker but carries a
+// distinct payload, so a single read of the value slot distinguishes three
+// states without a parallel initialized[] array: a real value (any other type),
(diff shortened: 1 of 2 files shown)


nooga (approved):
Thanks Matt — the heapEmptySlot sentinel is a clean way to fold the flag array away, and TypeUninitialized being internal-only with no other payload:1 usage rules out any collision. Merging.

PR #32, opened 2026-07-22
perf(vm): allocation-free Map/Set keys via a comparable struct

mparrett:
`hashKey` turned every Map/Set key into a string (`"s:"+content`, `"n:"+FormatFloat`, `"o:"+Sprintf("%p")`), so every set/get/has/delete/add allocated — and object keys paid `fmt.Sprintf`'s reflection.

## What this does

Replace the string key with a comparable value struct `mapKey{kind, bits, str}`: a tag plus a `uint64`, and for string keys a header aliasing the existing backing. Constructing one does no heap work; Go hashes/compares it field-wise, giving the same SameValueZero partitioning — NaN === NaN via a canonical bit pattern, +0 === -0 via bits 0, key-type disambiguation via the `kind` tag instead of the string prefix. Object keys hash on pointer identity as before (the Map/Set holds the key strongly and Go never moves heap objects). Number keys collapse identically (bit-equal iff shortest-decimal-equal); delete/re-insert order preserved (the stale order slot is the zero `mapKey`, which matches no entry).

**Design point for review:** the SameValueZero canonicalization (NaN via canonical bits, ±0 via bits 0), the pointer-identity object key and why it stays GC-safe, and the tradeoff below.

**Tradeoff:** `mapKey` is 32 B vs a 16 B string header, so maps hold ~16 B more per entry; `clear()`+regrow-heavy workloads allocate more in the bucket arrays but are still faster on wall-clock.

## Numbers

Local (M2): get/has-heavy map (build-once) ~12% with fewer GCs; object-key set churn (was `fmt.Sprintf`) ~13%; number-key churn ~5%. The `perf`-label A/B is authoritative.

## Verification

`TestScripts` green + `tests/scripts/map_set_key_hashing.ts` (key-type disambiguation, NaN/±0 canonicalization, object identity, delete/re-insert order); vm tests under `-race` clean; Test262 language 0 new failures (differential vs upstream/main control); Map/Set/WeakMap/WeakSet built-ins 0 new failures.

---

### DIFF
--- a/pkg/vm/value.go
+++ b/pkg/vm/value.go
@@ -291,10 +291,10 @@ type AsyncGeneratorObject GeneratorObject
 type MapObject struct {
 	Object
 	size       int
-	entries    map[string]Value // key -> value
-	keys       map[string]Value // key -> original key (for key iteration)
-	order      []string         // insertion order of keys (for deterministic iteration)
-	tombstones map[string]bool  // keys that were deleted but still in order (for live iteration)
+	entries    map[mapKey]Value // key -> value
+	keys       map[mapKey]Value // key -> original key (for key iteration)
+	order      []mapKey         // insertion order of keys (for deterministic iteration)
+	tombstones map[mapKey]bool  // keys that were deleted but still in order (for live iteration)
 	Properties *PlainObject     // User-defined properties on the Map object
 	prototype  Value            // Map prototype
 }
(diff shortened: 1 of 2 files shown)


nooga (approved):
Thanks Matt — the tagged mapKey struct correctly preserves SameValueZero (NaN canonicalization, +0/-0 folding, pointer identity for objects), and the regression tests cover the tricky cases well. One tiny non-blocking note: GetEntryAt/ForEach dropped the old fallback string-key synthesis on a lookup miss — reasonable given the stated invariant, just flagging it as slightly less defensive than before. Merging.

PR #33, opened 2026-07-22
perf(vm): pool the native-reentry sentinel register slice

mparrett:
Every native→JS reentry allocated a fresh `make([]Value, 1)` to hold the sentinel frame's result — one heap slice per callback invocation (every Array.map/forEach/filter/reduce/sort element, Promise executor, JSON reviver).

## What this does

Reentries nest strictly LIFO, so a small free-list recycles these 1-element buffers: it grows to the max nesting depth once and then hands out reused slices. `put` clears the slot so a pooled buffer doesn't retain the last result. The result is read (via `run()`'s return or the native-direct path) before the deferred `put`, so clearing is safe. Self-contained in `pkg/vm/vm.go` + `vm_init.go`.

**Design point for review:** the LIFO free-list invariant (reentries nest strictly) and that `put` clears the slot for GC hygiene.

## Numbers

Local (M2, min of 6): a forEach+map+reduce+filter loop 1389 → 1184 ms (~15%) with ~18% fewer GC cycles (320 → 264). The `perf`-label A/B is authoritative.

## Verification

`TestScripts` green; vm tests under `-race` clean; nested/recursive callback correctness (map-in-map, sort comparator, deep recursion) verified. Test262 language 0 new failures (differential vs upstream/main control); callback-heavy built-ins 0 new failures.

---

### DIFF
--- a/pkg/vm/vm.go
+++ b/pkg/vm/vm.go
@@ -160,6 +160,14 @@ type VM struct {
 	registerStack [RegFileSize * MaxFrames]Value
 	nextRegSlot   int // Points to the next available slot in registerStack
 
+	// sentinelRegPool is a LIFO free-list of 1-element register slices reused as
+	// the sentinel frame's result holder on native->JS reentry
+	// (executeUserFunctionSafe / ...WithNewTarget). Reentries nest strictly, so
+	// the pool grows to the max nesting depth and then recycles - eliminating a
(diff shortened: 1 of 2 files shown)


nooga (approved):
Thanks Matt — the LIFO pooling with defer-based return, and the confirmation that the sentinel frame has no closure so nothing can retain a pointer into the pooled slice, checks out. Merging.

PR #34, opened 2026-07-22
perf(vm): allocation-free string relational comparison for BMP strings

mparrett:
`<`, `>`, `<=`, `>=` on strings called `compareStringsUTF16`, which materializes both operands as `[]uint16` (`StringToUTF16` does make + `[]byte(s)`) — ~4 allocations per comparison, on the hot path for any string-sorting or relational loop. The equality path already guarded this; the relational path did not.

## What this does

Add `compareStrings`, which uses Go's native byte comparison when neither operand contains an astral character or lone surrogate — exactly the cases where UTF-8 byte order diverges from UTF-16 code-unit order (astral chars sort as high surrogates 0xD800+ in UTF-16 but above all BMP in code-point order). For everything else byte order == UTF-16 order, including the shorter-is-less tiebreak. Astral/surrogate content still falls back to full UTF-16, so ordering stays spec-correct. Both relational sites (pre- and post-`toPrimitive`) routed through it.

**Design point for review:** why byte order == UTF-16 order for BMP (and only diverges on astral), mirroring the existing equality guard.

## Numbers

Local (M2, min of 5): a string-relational loop 647 → 479 ms (~26%) with GC cycles dropping 39 → 1. The `perf`-label A/B is authoritative.

## Verification

`TestScripts` + unit tests green; correctness across BMP, multibyte-BMP (accented/CJK), prefix/length, and the astral edge (`U+E000 > U+10428` in UTF-16 order — which byte/code-point order gets wrong) verified. Test262 language 0 new failures across all four relational operators (differential vs upstream/main control); built-ins/String 0 new failures.

---

### DIFF
--- a/pkg/vm/vm.go
+++ b/pkg/vm/vm.go
@@ -2296,7 +2296,7 @@ startExecution:
 					// String comparison (lexicographic by UTF-16 code units per ECMAScript)
 					leftStr := leftVal.ToString()
 					rightStr := rightVal.ToString()
-					cmp := compareStringsUTF16(leftStr, rightStr)
+					cmp := compareStrings(leftStr, rightStr)
 					switch opcode {
 					case OpGreater:
 						result = cmp > 0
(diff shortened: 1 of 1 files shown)


nooga (approved):
Thanks Matt — reusing the existing stringNeedsUTF16Comparison guard rather than inventing new astral-detection logic is exactly the right move here. Merging.

PR #38, opened 2026-07-25
perf(vm): number fast paths, IsObject range check, OpSetIndex alloc fix

mparrett:
First slice of the remaining Tier-1 dispatch-loop work (comparisons / integer `%` / return-path cleanup follow in a stacked PR). Three related hot-path changes plus the VM bench harness that measures them.

> **Overlap with #20:** the leading commit is the same harness already open as #20. Merge #20 first and this rebases to three commits; or take this PR and close #20 as superseded.

## What this does

- **Number fast paths for `+ − × ÷ %` and `ToInteger`.** When both operands (or the input) are already numbers — the dominant case — take a direct path instead of the general boxing/coercion route.
- **`IsObject` OR-chain → range check.** Object-like `ValueType`s occupy a contiguous enum range; a single range check replaces the OR-chain. Hot path calls are predominantly on non-objects (numbers), where the old chain was worst-case. `TestIsObjectExhaustive` locks the equivalence.
- **`OpSetIndex` stops allocating on the common write.** Guard the prototype-chain setter walk so own data properties / dense array elements store in place. `array_inherited_index_setter.ts` pins that inherited setters still fire.

## Numbers

Local (Apple M2; also see the recent three-way shootout of the full Tier-1 stack):

| Bench | Effect |
|---|---|
| `BenchmarkSetIndex` | ~73% less B/op, ~26% fewer allocs |
| `BenchmarkIsObject` | large (range check vs OR-chain on non-objects) |
| `BenchmarkAdd` / `BenchmarkArith` / `BenchmarkToInteger` | covered by the number fast paths |

The `perf` label's same-runner A/B is the authoritative check.

## Verification

- `TestScripts` green; new/updated scripts pin inherited-setter and harness workloads.
- Test262 language: 0 new failures (certified on the merged Tier-1 stack in the local shootout: identical 23222/426 vs upstream tip).
- No semantics changes.

## Stack

Follow-up: **`perf/vm-dispatch-extras`** (comparisons, integer `%`, non-allocating finally check, dead-code cleanup) stacks on this branch.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pkg/vm/isobject_bench_test.go
+++ b/pkg/vm/isobject_bench_test.go
@@ -1,0 +1,62 @@
+package vm
+
+import "testing"
+
+// isObjectMix models the hot-path distribution: IsObject is called predominantly
+// on non-object values (numbers in arithmetic/ToInteger guards), which is the
+// OR-chain's worst case (all comparisons fail before returning false).
+func isObjectMix() []Value {
+	vs := make([]Value, 0, 32)
+	for i := 0; i < 24; i++ {
+		vs = append(vs, IntegerValue(int32(i))) // 24 numbers (typ far below the object range)
+	}
+	vs = append(vs,
+		NewString("x"), BooleanValue(true), Undefined, Null, // more non-objects
+		Value{typ: TypeObject}, Value{typ: TypeArray},
+		Value{typ: TypeProxy}, Value{typ: TypeMap}, // a few objects
+	)
+	return vs
+}
+
+func BenchmarkIsObject(b *testing.B) {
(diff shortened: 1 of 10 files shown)


nooga (approved):
Thanks Matt — verified the IsObject range check against the enum (TypeObject..TypeProxy is genuinely contiguous, locked in by TestIsObjectExhaustive) and confirmed the OpSetIndex &registers[baseReg] swap is safe since opSetProp/opSetPropSymbol never write through that pointer. Merging.

PR #40, opened 2026-07-25
perf(vm): numeric fast path for comparisons + integer % [1/4 of #39]

mparrett:
Part 1 of the #39 split.

Collapses the per-opcode number guards into one fast path shared by all 13 arithmetic and comparison opcodes, reading the type tags directly rather than through `ToFloat()` (too large to inline). Comparisons previously had no fast path at all and paid the full `ToPrimitive`/`helperCallDepth`/unwinding bookkeeping on every loop condition.

**Semantics.** `ToPrimitive` is the identity on numbers, so the fast path computes what the general path would. NaN needs no special casing: Go's float comparisons are already false in all four relational directions, and `l == r` is false for NaN, matching `StrictlyEquals`. For two numbers abstract and strict equality coincide, so `OpEqual`/`OpStrictEqual` share an arm.

**Integer `%`.** `math.Mod` is a software frexp/ldexp loop, ~7% of the recursion benchmark's profile. Go's truncated `int64` remainder takes the dividend's sign, exactly matching JS, so the integer path is used only when both operands round-trip exactly through `int64` (±2^53), the divisor is nonzero, and the result isn't a signed zero — JS requires `-0` for a negative dividend with zero remainder, which the integer path would lose. Everything else falls through to `math.Mod`. `tests/scripts/remainder_semantics.ts` pins these, including `1 / (-7 % 7) === -Infinity`.

The third commit drops the per-op number guards that the unified path made unreachable — the cleanup nit from your review.

The inner switch carries a `default` arm that degrades to the general path. It's unreachable today (the enclosing `case` list and the switch cover the same 13 opcodes), but without it, adding an opcode to that list without a fast-path arm would `continue` with `destReg` never written — a silently stale register, which is a worse failure than the slow path.

| | branch | contents |
|---|---|---|

### DIFF
--- a/pkg/vm/vm.go
+++ b/pkg/vm/vm.go
@@ -2018,18 +2018,6 @@ startExecution:
 			// Type checking specific to operation groups
 			switch opcode {
 			case OpAdd:
-				// Fast path: both operands already numbers. Skips two non-inlinable
-				// vm.toPrimitive calls (each with helperCallDepth/unwinding/handler
-				// bookkeeping) plus the string/BigInt/Symbol branches. Semantically
-				// identical to the slow path: toPrimitive is the identity for
-				// numbers, and the numeric branch computes
-				// Number(lhs.ToFloat() + rhs.ToFloat()).
-				if lt, rt := leftVal.typ, rightVal.typ; (lt == TypeIntegerNumber || lt == TypeFloatNumber) &&
-					(rt == TypeIntegerNumber || rt == TypeFloatNumber) {
-					registers[destReg] = Number(leftVal.ToFloat() + rightVal.ToFloat())
-					continue
-				}
-
 				// JS semantics: ToPrimitive on both first (for string check),
 				// then if either is String → concatenate ToString(lhs)+ToString(rhs);
 				// else ToNumeric on both; if both BigInt → BigInt add; else Number add.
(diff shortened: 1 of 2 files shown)


mparrett:

| benchmark | Δ |
|---|---|
| `FibPlaceholderRun` | **−23.5%** |
| `Arith` | **−21.5%** |
| `MatrixMult` | **−19.8%** |
| `Add` | **−18.8%** |
| `SetIndex` | −4.0% |

**This is the one clear win of the six open perf PRs.** The pattern is coherent with what the change claims: the four numeric-heavy workloads all move together by roughly a fifth, while `SetIndex` — array writes rather than arithmetic — barely moves. That internal contrast is most of why I believe it.

One caveat in your favour: the four-round session run *understated* this at −13 to −18%, because the calibration anchor took a ~4% transient excursion during two of this commit's four rounds and normalising by it ate part of the win. The targeted A/B above avoids that.

**Floors from null controls in the same run:** two commits compiling to byte-identical binaries measure up to **3.7%** apart on `./tests`. A layout control (real code added, never executed) is no worse — so anything under ~4% here is not attributable to the change. Full write-up and raw data: [`perf-session-remeasure-results.md`](https://github.com/mparrett/project-docs/blob/main/docs/paserati/perf-session-remeasure-results.md).

nooga (approved):

PR #41, opened 2026-07-25
perf(vm): non-allocating finally-handler check on returns [2/4 of #39]

mparrett:
Part 2 of the #39 split.

`findAllExceptionHandlers` allocates a `[]*ExceptionHandler` on every `OpReturn`/`OpReturnUndefined`, where the common case is an empty exception table and the result is immediately reduced to "first handler that is a finally". `findPendingHandler` fuses that filter into the scan and returns the handler directly.

It's an exact refactor of the selection logic: same `frameCount`/`closure` guards, same exception table, same iteration order, same `pc >= TryStart && pc < TryEnd` predicate. The `includeIterCleanup` flag covers the two call sites that also accept `IsIteratorCleanup`. `findAllExceptionHandlers` is unchanged and still used by the unwinding and generator paths.

Also reorders `vm.frameCount == 1 && function != nil && function.Name == "<script>"` so the plain load short-circuits before the string compare, which otherwise ran on every function return.

| | branch | contents |
|---|---|---|

### DIFF
--- a/pkg/vm/exceptions.go
+++ b/pkg/vm/exceptions.go
@@ -42,6 +42,30 @@ func (vm *VM) findAllExceptionHandlers(pc int) []*ExceptionHandler {
 	return handlers
 }
 
+// findPendingHandler returns the first handler in exception-table order that
+// covers pc and is a finally block (or, when includeIterCleanup is set, an
+// iterator-cleanup block). It selects the same handler the callers of
+// findAllExceptionHandlers used to pick, but without allocating a slice —
+// this runs on every OpReturn/OpReturnUndefined, where the common case is an
+// empty table.
+func (vm *VM) findPendingHandler(pc int, includeIterCleanup bool) *ExceptionHandler {
+	if vm.frameCount == 0 {
+		return nil
+	}
+	frame := &vm.frames[vm.frameCount-1]
+	if frame.closure == nil {
+		return nil
+	}
+	table := frame.closure.Fn.Chunk.ExceptionTable
+	for i := range table {
+		h := &table[i]
+		if pc >= h.TryStart && pc < h.TryEnd && (h.IsFinally || (includeIterCleanup && h.IsIteratorCleanup)) {
+			return h
+		}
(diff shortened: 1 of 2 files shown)


mparrett:
**Regression on the arithmetic workloads.** Measured on a dedicated `c7a.2xlarge` (EPYC 9R14), go1.26.0, against the shared merge base `3167412e85c2`. Targeted A/B: alternating launches, `b.N` pinned at 8, median of 5–8 launches per arm. Negative is faster.

| benchmark | Δ |
|---|---|
| `Arith` | **+29.3%** |
| `Add` | **+26.2%** |
| `MatrixMult` | +3.9% |
| `FibPlaceholderRun` | −0.5% |
| `SetIndex` | −6.0% |

`Arith` was re-run with true per-launch interleaving because the size of it was surprising: **+29.3%** interleaved against +29.5% grouped, with a 1.0% spread within each arm. It reproduces.

The shape is odd and might point at the cause: `Arith` and `Add` regress hard, `Fib` and `MatrixMult` do not, and `SetIndex` improves. If the finally-handler check moved onto a path that tight arithmetic loops take but recursion and matrix work do not, that would fit — worth a look before this lands.

mparrett:
**Correcting my verdict here: the +29.3% is real, and it is not this change.**

The exposure runs the wrong way. `Fib` executes the `OpReturn` path this PR
changes about **five million times** per iteration and measured **−0.5%**.
`Arith` and `Add`, which execute roughly **two** returns against millions of loop
iterations, measured +29.3% and +26.2%. The benchmarks that regressed are the
ones that barely touch the changed code.

What is left is code layout — this change re-encodes **35.6%** of `(*VM).run`'s
65,397 instructions through spill-slot reassignment. Details and the missing
control in #52.

The non-allocating handler check still looks right on its merits, and it has no
demonstrated cost or benefit on this suite. I would leave this open and
unmeasured rather than close it as a regression.

nooga (approved):

PR #42, opened 2026-07-25
perf(vm): drop dead debug blocks and redundant IsNaN guards [3/4 of #39]

mparrett:
Part 3 of the #39 split.

Two removals, both provably dead:

- The `NUCLEAR DEBUG` blocks in `OpGetGlobal`/`OpSetGlobal` — they build a function name and then test it against an empty `if` body.
- The explicit `IsNaN` guard in the relational slow path. Per ECMAScript a NaN operand makes every relational comparison false, which is exactly Go's float semantics, so the guard only restates what the comparison already does.

**On your bounds-guard concern:** the `ip >= len(code)` check is kept. The commit that removed it in #39 isn't part of this split, so `main` never lost it — this branch retains it and adds a comment recording why it stays (unreachable after the fall-off-end handler above, kept so a future compiler bug surfaces as a graceful `runtimeError` instead of a Go panic on `code[ip]`).

| | branch | contents |
|---|---|---|

### DIFF
--- a/pkg/vm/function.go
+++ b/pkg/vm/function.go
@@ -172,8 +172,6 @@ func (fn *FunctionObject) GetOrCreatePrototype() Value {
 // GetOrCreatePrototypeWithVM lazily creates and returns the function's prototype property,
 // using the VM's prototypes for proper inheritance chain setup.
 func (fn *FunctionObject) GetOrCreatePrototypeWithVM(vm *VM) Value {
-	// NUCLEAR DEBUG
-
 	// Ensure Properties object exists
 	if fn.Properties == nil {
 		fn.Properties = NewObject(Undefined).AsPlainObject()
@@ -183,7 +181,6 @@ func (fn *FunctionObject) GetOrCreatePrototypeWithVM(vm *VM) Value {
 	if proto, exists := fn.Properties.GetOwn("prototype"); exists {
 		return proto
 	}
-
 
 	// Determine the correct prototype parent based on function type
 	var prototypeParent Value = DefaultObjectPrototype
--- a/pkg/vm/vm.go
+++ b/pkg/vm/vm.go
@@ -1325,10 +1325,13 @@ startExecution:
 			}
 		}
 
-		// Defensive: unreachable after the fall-off-end handler above (which
-		// already covers ip >= len(code) with continue/return). Kept as a
-		// cheap belt-and-suspenders check so a future compiler bug surfaces
-		// as a graceful runtimeError rather than a Go panic on code[ip].
+		// Defensive backstop, unreachable today: the fall-off-end handler
+		// directly above absorbs every ip >= len(code), either popping the
+		// frame or returning, and nothing between the two touches ip or code.
+		// Kept so that if that handler's coverage ever narrows, an out-of-range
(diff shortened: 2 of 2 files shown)


mparrett:

| benchmark | Δ |
|---|---|
| `Add` | **+11.3%** |
| `Arith` | +2.3% |
| `FibPlaceholderRun` | +1.7% |
| `SetIndex` | −0.2% |
| `MatrixMult` | −0.0% |

**Mostly neutral, with one real regression on `Add`** at +11.3% — comfortably outside the ~4% floor measured in the same run. The other four are inside it and should be read as no change.

Dropping dead debug blocks and redundant guards should not cost anything, so +11.3% on one workload suggests the removal shifted something on the `Add` path specifically — either a guard that was doing useful work, or code layout. Worth a look given the change is meant to be free.

**Floors from null controls in the same run:** two commits compiling to byte-identical binaries measure up to **3.7%** apart on `./tests`, and a layout control is no worse — so anything under ~4% is not attributable. These are far outside that. Full write-up and raw data available on request.

mparrett:
Following up on my own numbers above with evidence that did not exist when I posted them. Two corrections: the floor I compared against was the wrong one, and I had missed that this diff touches `Add`'s inner loop.

**The floor.** I called +11.3% "comfortably outside the ~4% floor". That floor came from null controls in the same session — commits compiling to byte-identical binaries — and it is a corpus-wide figure. `Add` is noisier than the corpus:

- Across the 96 snapshots on the `perf-data` branch of my fork, comparing same-build, same-machine pairs whose true difference is zero by construction, `BenchmarkAdd`'s worst spread is **5.16%** and its median is 3.21% (recomputed 2026-08-12). Reproducible from that branch alone: group snapshots by `provenance.build_hash` and machine key, then take the spread of `ratio_to_anchor` within each group holding more than one commit.
- In a three-point bisect probe on a dedicated box (2026-08-06), `Add` had the worst round-to-round agreement of any micro measured: **9.46pp**, against 0.02pp for `IsObject` and 0.06pp for `NavierStokes`. Its deltas were recorded there as "not resolvable at N=9". That run is not published; I can share the raw output if it is worth anything to you.

So +11.3% clears the same-build floor but sits inside the range `Add` moves between rounds on its own.

**The inner loop.** `Add` is also the one benchmark in that table whose hot path this diff changes. `bench_add.ts` drives `for (let i = 0; i < ITERS; i++)`, and that condition compiles to a single `OpLess` beside the four `OpAdd`s, so it runs the `math.IsNaN` branch removed here two million times per benchmark iteration. `Arith`, `SetIndex`, `MatrixMult` and `FibPlaceholderRun` never reach it.

That points in two directions at once. A real effect on `Add` is plausible in a way it would not be for the other four. But the sign is wrong: dropping a never-taken branch out of a hot loop should not cost 11%, so a real number here would indict code layout rather than the deleted work.

Recommendation: merge. The regression is not established by this evidence, and the one benchmark that could plausibly have moved should have moved the other way. If it is worth resolving properly, it needs the samples retained and compared as distributions — the standing point in #22 — rather than another min-then-median-of-N A/B, which is the method that already failed to resolve `Add` at N=9.

nooga (approved):

PR #43, opened 2026-07-25
perf(vm): make arrayIndexAccessorSeen an atomic.Bool [4/4 of #39]

mparrett:
Part 4 of the #39 split — the `arrayIndexAccessorSeen` concern from your review.

The latch is now an `atomic.Bool`. As you noted it was fail-safe (a stale `false` just restores the slow path), but it was still a data race under `-race` or with concurrent VM instances; `Store`/`Load` make the publication well-defined.

`noteAccessorKey` keeps its load-then-store shape rather than a CAS — the only transition is `false → true`, so a redundant store is harmless.

`go test -race ./pkg/vm/` passes.

| | branch | contents |
|---|---|---|

### DIFF
--- a/pkg/vm/object.go
+++ b/pkg/vm/object.go
@@ -670,7 +670,11 @@ func (o *PlainObject) DefineOwnProperty(name string, value Value, writable *bool
 // rather than per-realm to keep the accessor-definition choke points VM-free; the
 // only cost of the coarser scope is that one realm defining such an accessor also
 // disables the fast path for others, which is negligible in practice.
-var arrayIndexAccessorSeen bool
+//
+// atomic.Bool so concurrent VMs / -race builds don't data-race the latch. A
+// torn false→true under a plain bool would be fail-safe (restore slow path) but
+// still races; Store/Load make the publication well-defined.
+var arrayIndexAccessorSeen atomic.Bool
 
 // isCanonicalArrayIndexKey reports whether name is a canonical array-index string
 // ("0", or a digit-string with no leading zero, within the 0..2^32-2 index range) —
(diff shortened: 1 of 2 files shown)


mparrett:
**This is a substantial regression, not a neutral cleanup.** Measured on a dedicated `c7a.2xlarge` (EPYC 9R14), go1.26.0, against the shared merge base `3167412e85c2`. Targeted A/B: alternating launches, `b.N` pinned at 8, median of 5–8 launches per arm. Negative is faster.

| benchmark | Δ |
|---|---|
| `MatrixMult` | **+33.2%** |
| `FibPlaceholderRun` | **+30.9%** |
| `Arith` | +10.4% |
| `Add` | +5.7% |
| `SetIndex` | −1.1% |

Alternating base-vs-head launches on `Fib`, ten each:

```
base  880  909  911  911  913  913  914  916  918  922  926   ms
head 1199 1200 1200 1201 1208 1218 1218 1219 1222 1225 1234   ms
```

No overlap between the two distributions, and each arm is tight (2.9–5.2% spread). This reproduces across every launch.

The mechanism is named in the commit message: *"nothing stops the compiler caching the load across an `arr[i] = v` loop."* That is exactly right, and it is the cost — making the flag an `atomic.Bool` forces a memory read where the compiler previously hoisted it out of the loop. The race is real and worth fixing, but not at this price.

Worth considering: the flag is only ever set once and never cleared, and `Array.prototype` is per-realm. A cheaper shape might be a plain read guarded by a one-time initialisation, or hoisting the check out of the hot path entirely, rather than an atomic load per access.

mparrett:
**Correcting my verdict here: the +30.9% is real, and it is not this change.**

The regressing benchmark is `Fib`, which compiles `scripts/factorial.ts` — a
fixture containing **zero `OpSetIndex` instructions**. The load this PR makes
atomic sits inside the `OpSetIndex` handler, so it cannot execute once in the
benchmark that moved. `SetIndex`, which executes that guarded load 200,256 times
per iteration, measured **−1.1%**.

So the un-hoistable-atomic-load reading I gave predicts the opposite ranking to
the one measured.

What is left is code layout. This change alters `(*VM).run`'s size by **zero
bytes** yet re-encodes **19.7%** of its 65,397 instructions: spill slots shift
(`151(RSP)` → `159(RSP)`) across a 261 KB function. Details and the missing
control in #52.

I would leave this open and unmeasured rather than close it as a regression.

nooga (approved):

Issue #44, opened 2026-07-25
VM: recovered panics are reported as successful runs (silent exit 0)


mparrett:
A Go panic anywhere inside the interpreter loop is recovered, discarded, and reported to the caller as a **successful** run. The process prints nothing further and exits 0.

## Mechanism

`VM.run` installs a blanket recover (`pkg/vm/vm.go:1250-1259`):

```go
func (vm *VM) run() (InterpretResult, Value) {
	defer func() {
		if r := recover(); r != nil {
			if debugVM {
				fmt.Printf("[PANIC] Recovered panic in VM.run(): %v\n", r)
				debug.PrintStack()
			}
		}
	}()
```

`debugVM` is `false` in normal builds, so the panic value and stack are dropped on the floor. Because the returns are unnamed, the recovered function then yields the zero values — and `InterpretOK` is the first `iota` in the `InterpretResult` block (`pkg/vm/vm.go:389-395`), so zero *is* `InterpretOK`.

The caller therefore cannot distinguish a crash from a clean run: it sees `InterpretOK` and an `undefined` result.

## Reproduction

Any panicking native will do. Before the BigInt fix, this was a one-liner:

```js
var b = 10n;
console.log("before");
console.log(b.toString());   // panics inside the native
console.log("after");
```

```
$ ./paserati --no-typecheck repro.js
before
$ echo $?
0
```

No error, no stack, no non-zero exit — `after` simply never runs.

## Why it matters

- **It hides bugs for as long as nobody looks.** The BigInt breakage meant *every* `BigInt.prototype` method was unusable on a primitive receiver, and it presented as "the script stops early."
- **It is invisible to a test harness.** Anything checking exit status or absence of stderr sees success. A panicking script looks like a passing one.
- **`debugVM` is the wrong gate.** It has to be flipped at compile time and turns on very heavy tracing, so nobody enables it just to find out why a script stopped.

## Possible directions

Deliberately not proposing a patch, since the right answer is a design call:

1. Convert the recovered panic into a runtime error — return `InterpretRuntimeError` with the panic value as the message, so it surfaces through the normal error path and sets a non-zero exit.
2. Re-panic after logging, so it fails loudly with a usable stack.
3. Keep recovering, but always write the panic value and stack to stderr regardless of `debugVM`, and return `InterpretRuntimeError`.

Any of the three would have made the BigInt bug obvious immediately. Option 1 seems most in keeping with the engine's existing `runtimeError` handling, but it is your call.

[...]

PR #45, opened 2026-07-25
builtins: fix BigInt prototype methods panicking on a primitive receiver

mparrett:
`10n.toString()` prints nothing and exits 0. So do `valueOf()`, `toLocaleString()`, and `BigInt(10n)` — every BigInt prototype method is unusable on a primitive receiver.

Two commits: the panic fix, and a separable radix fix that the first one makes reachable.

## 1. The panic

All four sites inlined the same object-wrapper unwrap:

```go
if v.Type() == vm.TypeBigInt {
    // For BigInt object wrappers, extract the primitive value from [[BigIntData]]
    if po := v.AsPlainObject(); po != nil {
```

`AsPlainObject` panics unless the value is `TypeObject` (`value.go:968-971`), and a BigInt object wrapper *is* `TypeObject` — never `TypeBigInt`. The guard admits exactly the values the body cannot handle: the wrapper branch was unreachable, and every primitive receiver panicked.

It reads as a silent stop rather than a crash because `VM.run`'s recover discards the panic and returns `InterpretOK` — filed separately as #44.

Replaced all four copies with one `thisBigIntValue` helper implementing ThisBigIntValue. Four hand-inlined copies of a subtle unwrap is what produced four identical bugs.

**The slot name matters here.** The spec calls it `[[BigIntData]]`, and that is what the old code looked up — but nothing in this repo ever writes that name. `Object(bigint)` stores `[[PrimitiveValue]]` (`object_init.go:944`), so searching for the spec name would have left the wrapper arm dead in a new place. The helper uses `[[PrimitiveValue]]` and checks the stored value's type, since Number/String/Boolean/Symbol wrappers share that slot.

## 2. The radix

`toString`'s radix was coerced with a bare `ToFloat` and clamped into 2..36, with a comment noting real JS throws. Now runs the package's existing `toIntegerOrInfinityWithVM` and throws `RangeError` outside 2..36.

Separable, but included because commit 1 is what makes these paths reachable — see the test262 accounting below. Happy to split it out if you'd rather.

## Verification

Primitive and wrapper receivers, all matching Node:

```
10n.toString()                          10
10n.toString(16)                        a
BigInt("-255").toString(16)             -ff
10n.valueOf()                           10n   (typeof bigint)
10n.toLocaleString()                    10
BigInt(10n)                             10n
Object(10n).toString()                  10
Object(10n).toLocaleString()            10
BigInt.prototype.valueOf.call(Object(10n))  10n
BigInt(Object(10n))                     10n
```

Non-BigInt receiver fallbacks are unchanged: `toString`/`toLocaleString` stringify, `valueOf` throws `TypeError`.

`tests/scripts/bigint_prototype_methods.ts` covers all of the above. With commit 1 reverted it fails with `"undefined"` — the script produces nothing at all.

`go build`/`vet`/`gofmt` clean; `TestScripts`, `pkg/vm`, `pkg/builtins`, `pkg/compiler` pass.

### test262

`built-ins/BigInt`: **41 → 41**, and `language`: 23222/23648 with zero per-test diffs.

The count is flat but the composition changed, so the per-test diff is worth stating plainly:

- **+1** `prototype/valueOf/cross-realm.js` — genuinely fixed by the wrapper unwrap.
- **−1** `prototype/toString/radix-tointegerorinfinity-throws-toprimitive-or-bigint.js` — this was a false pass. The panic ended the script before its assertion ran, and the harness scored the silence as green. It now fails for a real, pre-existing reason: `toIntegerOrInfinityWithVM` doesn't reject a BigInt returned from `ToPrimitive`. That helper is shared with Array, so fixing it belongs in its own change where the Array impact can be measured — deliberately not touched here.

Two other radix tests (`radix-err.js`, `radix-tointegerorinfinity-throws-symbol.js`) were false passes for the same reason; commit 2 turns those into genuine passes, which is why the total lands back at 41 rather than 38.

## Note on the test

It calls through `any`. The checker rejects property access on a bigint receiver outright ("property access is not supported on type bigint"), rejects unary `-` on bigint, and types a bigint literal as a number literal (`const b: bigint = 10n` fails). All pre-existing and unrelated to this PR — happy to file separately if you want them tracked.

### DIFF
--- a/pkg/builtins/bigint_init.go
+++ b/pkg/builtins/bigint_init.go
@@ -42,6 +42,35 @@ func (b *BigIntInitializer) InitTypes(ctx *TypeContext) error {
 	return ctx.DefineGlobal("BigInt", bigintCtorType)
 }
 
+// thisBigIntValue implements the spec's ThisBigIntValue: a primitive BigInt is
+// its own value, and a BigInt object wrapper yields its wrapped primitive.
+// Any other receiver reports false.
+//
+// The slot is spelled "[[PrimitiveValue]]" because that is what actually
+// creates BigInt wrappers here (object_init.go's Object(bigint) case); the
+// spec's name for it is [[BigIntData]]. Number/String/Boolean/Symbol wrappers
+// share the same slot name, so the TypeBigInt check on the stored value is
+// what keeps this from unwrapping one of those.
+//
+// This lives in one place on purpose. Each call site previously inlined the
+// unwrap guarded by `Type() == TypeBigInt` and then called AsPlainObject() on
+// it — but AsPlainObject panics unless the value is TypeObject, and a wrapper
+// is TypeObject, never TypeBigInt. So the wrapper branch was unreachable and
+// every primitive receiver panicked.
+func thisBigIntValue(v vm.Value) (vm.Value, bool) {
+	switch v.Type() {
+	case vm.TypeBigInt:
+		return v, true
+	case vm.TypeObject:
+		if po := v.AsPlainObject(); po != nil {
+			if data, exists := po.GetOwn("[[PrimitiveValue]]"); exists && data.Type() == vm.TypeBigInt {
+				return data, true
+			}
+		}
+	}
+	return vm.Undefined, false
+}
+
 func (b *BigIntInitializer) InitRuntime(ctx *RuntimeContext) error {
 	vmInstance := ctx.VM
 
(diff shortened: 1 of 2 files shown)


nooga (approved):
Verified locally: merges cleanly against current main, builds/vets clean. Reproduced the root-cause panic directly on main (10n.toString() silently swallows a panic and exits 0 without running subsequent code) and confirmed the fix resolves it. Confirmed [[PrimitiveValue]] (not the spec's [[BigIntData]]) is genuinely the slot object_init.go writes for wrappers - the fix targets the right key. Manually reproduced all outputs the PR claims (toString radix, wrapper valueOf, RangeError on bad radix). Test262 built-ins/BigInt diffed against a main baseline: +1 pass, 0 regressions locally.

PR #47, opened 2026-07-26
TypeScript conformance: strict defaults, TS2454, and using declarations

nooga:
Raises TypeScript conformance from **3,197 to 3,470 of 4,933 (64.8% → 70.3%)** against TypeScript **v6.0.3**, and adds 22 Test262 language passes.

The conformance checkout was six months stale on `main`; it is now pinned to release tag `v6.0.3`, which is what surfaced the first item below.

## `paserati-testtsc`: match TS 6.0 strict defaults

TS 6.0 made `strict` the default and regenerated its conformance baselines to match, so tests with no `// @strict` directive now expect TS2564 and TS2454. The runner still defaulted those gates off, which cost ~146 tests and read as a Paserati regression rather than a harness default. Multi-value directives (`// @strict: true, false`) now resolve to the variant the selected baseline actually carries.

## TS2454 definite assignment analysis

New `pkg/checker/definite_assignment.go`: a forward dataflow walk per function scope. Conditionals merge by intersection, with a terminating arm letting the other decide alone; short-circuiting operators treat their right side as maybe-not-run; loops read pre-loop state. Where flow is not modelled precisely — loops, try/catch, switch, class static blocks — every syntactic write inside is unioned in afterwards, so an unmodelled path can only *silence* a diagnostic, never invent one. Candidates touched from a nested function are dropped, as are `var`s merging with a pre-existing global.

Also adds the `let x!: T` assertion to the parser, which meant consolidating six copies of the declarator annotation parsing into `parseDeclaratorAnnotation`.

**+101 conformance tests, zero new failures across the suite.**

## `using` / `await using` declarations

Explicit resource management (ES2026), as a contextual keyword at statement position and in both for-head forms. Blocks containing `using` are rewritten into try/finally after type checking, so disposal reuses the existing abrupt-completion paths rather than adding a new unwinding mechanism. Reverse-order disposal, per-iteration disposal in `for-of`, loop-lifetime disposal in `for (using x = e;;)`, null/undefined as no-ops, and a spec `TypeError` when a resource has no dispose method. Adds `Symbol.asyncDispose`.

**usingDeclarations 43 → 63; Test262 `language/statements` +10.**

## Two things worth a look

- **Seven smoke tests changed.** They used the `let x: T;`-then-read idiom TS2454 rejects. I ran each through real `tsc` (`lib/tsc.js --strict`) first — it reports all of them and more — then added `!`, which preserves each test's subject and changes no runtime behaviour.
- **One conformance test regressed**, `expressions/functionCalls/typeArgumentInferenceWithObjectLiteral.ts`. Bisected to adding `asyncDispose` to the `Symbol` constructor *type*. It was only passing because we emit a spurious error on line 31 unrelated to the TS2345 its baseline expects. `Symbol.asyncDispose` is real ES2026 and belongs there; the spurious error is a separate pre-existing bug.

## Known gap

`using` at the **top level of a module** still fails when exports follow it (3 tests): the lowering moves later statements into the try block, putting exported bindings out of module scope. It surfaces as a compile error rather than silent mis-scoping, and is not a regression — `using` did not parse at all before. A proper fix needs declaration hoisting in the lowering.

## Verification

`go test ./tests -run TestScripts` and `go test ./pkg/...` green. Test262 `language` diffs +22 / -0 against baseline. New smoke tests cover definite assignment (positive and negative), `using` disposal semantics, and the for-head lookahead.


nooga:
## Review: solid compliance gains, but several confirmed correctness bugs before merge

Verified locally: full local TypeScript-conformance diff vs `main` is **+291 net** (327 new passes, 36 new failures), Test262 `language` is **+10 / 0 regressions**, and `using`/`await using` core disposal order (LIFO) and basic TS2454 cases work correctly. But the following are independently reproduced bugs, not style nits — recommend fixing before merge.

### Confirmed bugs

1. **TS2454 false negative via object-literal shorthand methods.** The analysis doesn't treat a shorthand method as a deferred-execution scope, so an assignment inside a method body is treated as always-executed in the enclosing scope:
   ```ts
   let x: number;
   const obj = { m() { x = 5; } };
   console.log(x + 1); // compiles clean, prints NaN at runtime
   ```

2. **`for (using x of xs)` skips the disposability guard.** A non-disposable per-iteration value runs the loop body first, then throws the wrong error:
   ```ts
   for (using x of [1, 2, 3]) { console.log(x); }
   // prints "1", THEN throws "TypeError: undefined is not a function"
   // should throw "TypeError: Object is not disposable." before any iteration
   ```

3. **TS1263 isn't enforced.** `!` combined with an initializer should be a compile error and isn't:
   ```ts
   const x!: number = 5; console.log(x); // compiles, prints 5
   ```

4. **Flow-narrowing overlay is keyed only by variable name, not by scope**, so a shadowed variable in an inner block corrupts an outer variable's narrowed type:
   ```ts
   function outer() {
     let y = "a";
     { let y = "b"; y = "c"; }
     let z: "a" = y; // false-positive TS2322 — outer y was never reassigned
   }
   ```

5. **TS2454 doesn't model switch fallthrough** — evaluates each case independently from pre-switch state:
   ```ts
   function f(x: number) {
     let y: number;
     switch (x) {
       case 1: y = 1;
       case 2: console.log(y); break; // false-positive TS2454; real tsc accepts this
       default: y = 0;
     }
   }
   ```

6. **Contextual keywords can't be used as `using` binding names**, though TypeScript allows it:
   ```ts
   using as = { [Symbol.dispose]() {} }; // PS1003, fails to parse
   ```

### Noted, not blocking on this PR

- **`await using` can lose a promise rejection** when the synthesized `finally` awaits during an active throw. Checked whether this is new: it isn't — `async function f(){ try { throw new Error("boom") } finally { await Promise.resolve() } } f().then(...)` already exhibits the same symptom on `main` today (neither callback fires; the runtime dumps the raw error object). This is a pre-existing try/finally+await bug that `using`'s lowering strategy inherits, not something introduced here — filing separately rather than folding an unrelated engine bug into this PR.
- **`SuppressedError` is entirely unimplemented** (0/22 in Test262 `built-ins/SuppressedError`). With two disposals that both throw plus a throwing block body, only the last error survives — the original body error and the first disposal error are silently discarded. This is the mechanism the spec added specifically to prevent that. Given the commit's own description ("blocks containing `using` are rewritten into the equivalent try/finally") — that's the same shape `SuppressedError` exists to fix, so this is a real gap, not a nit. Recommend scoping this as a separate follow-up (new builtin: constructor, prototype, `error`/`suppressed` properties, realm + type wiring, ~22 Test262 tests) rather than growing this already-3,945-line PR further — but at minimum this PR should land a test pinning the current (lossy) disposal-error behavior so the gap is documented rather than silent.

### Merge mechanics

- Onto current `main` this isn't a clean rebase: PR47 independently added union-of-object-types destructuring support in the same function PR14 (already merged) touched. Needs deliberate reconciliation — PR14's `destructureType` annotation-priority logic has to apply to PR47's new union branch too, not just the object/array branches, or the annotation-ignored bug PR14 fixed comes back for union-typed destructuring initializers.
- PR47 also supersedes PR14's ad-hoc destructuring-assignability error wording with real TS2322 wording, which breaks `tests/scripts/ts_destructuring_annotation_non_literal.ts`'s hardcoded expected substring (the check itself still fires correctly — just needs the test's expected string updated as part of this rebase).

Working on fixes for items 1–6 now.

nooga:
## Correction to finding #1 (TS2454 shorthand-method false negative)

On investigation this is **not** a shorthand-method-specific gap, and I'm not fixing it as filed. Tracing it down:

- Object-literal shorthand methods parse into `MethodDefinition{Value: *FunctionLiteral}`, not `*parser.ShorthandMethod`. `ShorthandMethod`'s only constructor (`transformShorthandMethodWithDestructuring`) has no call site anywhere in the codebase — it's dead code for this syntax (worth a separate cleanup pass, unrelated to this PR).
- `MethodDefinition.Value` being a plain `*FunctionLiteral` means the definite-assignment analysis's existing `nestedFunctionBody` recognition and the `walk()` exclusion **already** treat a shorthand method's body as a nested scope, the same as a `function`/arrow expression's body — there was never a missing case to add.
- Confirmed with a controlled A/B: I patched `ShorthandMethod` into both call sites, built two binaries from git-stashed before/after state, and ran the original repro against both. Identical output on both — the patch was a no-op, which is exactly what you'd expect if the real node type was already handled.
- The "compiles clean, prints NaN" symptom reproduces identically for a plain arrow function (`const f = () => { x = 5; }`) on unmodified `main` too. It's the module's own documented, intentional behavior (`definite_assignment.go`'s header: "Candidates touched from a nested function are dropped entirely, since their flow is unknowable here") applied uniformly across every closure form — not something PR47 introduced or something specific to methods.

Net: not a bug to fix here. If you want TS2454 to track whether a stored closure has actually run before a later read (which real `tsc` does more precisely), that's a real feature gap, but it's a redesign of the whole nested-closure-disqualification strategy, not a targeted fix — worth its own issue if you want it, separate from this PR.

Continuing with the remaining confirmed findings (switch fallthrough, flow-narrowing scope corruption).

nooga:
## Correction to finding #4 (flow-narrowing overlay scope corruption)

Same pattern as finding #1 — also not fixing this one, and also not a real bug given how the analysis is actually wired up.

The repro claims the bug is *shadowing*-specific:

```ts
function outer() {
  let y = "a";
  { let y = "b"; y = "c"; }
  let z: "a" = y; // claimed: false-positive TS2322
}
```

Isolating test — same shape, **no shadowing at all**, just an unrelated empty block:

```ts
function outer() {
  let y = "a";
  { console.log("hi"); }
  let z: "a" = y; // also errors, identically
}
```

Both produce the exact same `TS2322`. That rules out shadowing/scope-crossing as the cause: it's `checker.go`'s block-statement handling. Every block (bare `{}`, `if`/`for`/`while` body, function body) gets its own fresh `flowNarrowState`, and — this is the load-bearing part — the *enclosing* sequence's loop invalidates its own overlay entries for any statement that isn't a plain `let`/`const`/`var`/expression-statement (the `default:` branch), and a block/if/for/function statement is never one of those. So the outer scope's narrowing for `y` is already gone *before* the nested block even runs, shadowing or not. This is `flow_narrowing.go`'s own documented, intentional design ("any other statement... invalidates everything tracked so far") — not a corruption bug, and not reachable through the current caller structure regardless of same-name shadowing.

Net: this is pre-existing, documented conservatism (paserati doesn't attempt to track narrowing across *any* nested block, matching its stated non-goal of full control-flow analysis), not something PR47 introduced or something actionable as a targeted fix here. Not touching it.

That leaves 4 of the original 6 confirmed bugs fixed and landed (the `for (using x of xs)` disposability guard, the contextual-keyword `using` binding names, TS1263, and TS2454's switch-fallthrough false positive) — commits incoming on the branch.

PR #55, opened 2026-08-15
fix(string,regexp): index by UTF-16 code unit, not UTF-8 byte

### LINKED ISSUE #54: String and RegExp methods index by UTF-8 byte, not UTF-16 code unit
String and RegExp methods index by UTF-8 byte where ECMAScript indexes by UTF-16 code unit. Any string holding one non-ASCII character makes them disagree with `length`, which is already correct — so the same string can report `length === 7` and return a 4-unit substring of its last 3 characters.

Reproduced on `main` at `3167412e`.

```js
var s = "abcédef";          // 7 code units, 8 UTF-8 bytes; 'd' is at index 4

s.length                    // 7   correct
s.charAt(4)                 // "d" correct
s.substring(4)              // expected "def"
                            // actual: 4 chars starting with code 169, é's UTF-8 continuation byte
s.indexOf("d")              // expected 4, actual 5
s.search(/d/)               // expected 4, actual 5
s.match(/d/).index          // expected 4, actual 5

var re = /d/g; re.exec(s);
re.lastIndex                // expected 5, actual 6
```

Walking `/./g` over that string visits `0,1,2,3,5,6,7`: it skips an index that exists and reports one that does not.

## Why it survived

`length`, `charAt`, `at`, `codePointAt`, `includes`, `padStart` and `split("")` go through `vm.StringToUTF16` and are right — I checked each against the same string. The rest take their offsets straight into the backing Go string, which is UTF-8. The two coordinate systems agree on ASCII and nowhere else, so everything looks fine until a string carries one accented character.

`lastIndex` is the load-bearing case: it is both JS-visible and the engine's own cursor, so it was storing a byte offset in a property the spec defines in code units. A caller reading it back cannot use it to slice the string it came from.

## Affected

- `String.prototype`: `substring`, `slice`, `substr`, `indexOf`, `lastIndexOf`, `startsWith`, `endsWith`
- Match positions: `search`, `match().index`, `exec().index`, `matchAll`, the offset passed to a `replace` callback, and `lastIndex`

Returned strings are correct throughout; the defect is confined to offsets.

## Effect on Octane RegExp

The workload fails its own checksum on `main`, and this bug is part of why. `computeInputVariants` rebuilds a string around one rotated character using `substring`, so every variant of a non-ASCII input comes out short. Scoring each of the twelve blocks separately against V8:

| block | V8 | `main` | `main` + this fix |
|---|---|---|---|
| 0 | 286139 | 286145 | 286145 |
| 1 | 1212496 | 1212189 | 1212495 |
| 2 | 5429 | 5430 | 5430 |
| 6 | 77823 | 77956 | 77823 |

The other eight blocks already agree. So the fix accounts for all of block 6 and all but 1 of block 1's drift, and blocks 0 and 2 are a separate cause I have not chased to the bottom — `\s` and `.` do not carry their ECMAScript meanings under RE2, which is its own bug. **The checksum still fails after this fix**; it fails by less.

## Fix shape

Convert at the boundary rather than holding strings as `[]uint16`: `utf16Len`, `utf16ToByteOffset` and `byteToUTF16Offset`, so the ASCII path stays allocation-free and byte-wise searches still run through `strings.Index` with their results mapped back before they reach the caller.

Converting `exec()`'s `index` is not sufficient on its own. `Symbol.replace` reads that property back and then uses it to slice `str`, so the same number was serving as a JS index and a byte offset; the two have to become separate variables or the fix trades one bug for another.

## Measured on `main`

I have this as two commits that cherry-pick onto `3167412e` cleanly, with smoke tests:

- `built-ins/RegExp`: 851 → 849 failing. Two newly passing, `prototype/exec/u-lastindex-value.js` and `S15.10.2.7_A2_T1.js`, no regressions.
- `built-ins/String`: 87 failing, unchanged, identical set.
- `language` against `baseline_language.txt`: no attributable change. A batch run of that suite is not repeatable to better than a few tests here — I saw net 0, −1 and −3 across runs, on `main` with and without the patch alike — so I would not read a small number either way as signal.

I'll open it as a PR unless you would rather review the String and RegExp halves separately.

mparrett:
Fixes #54.

Seven `String.prototype` methods and every match position the regex path
reports took their offsets straight into the backing Go string, which is
UTF-8, while `length`, `charAt` and `split("")` go through
`vm.StringToUTF16` and report code units. The two coordinate systems agree
on ASCII and nowhere else, so one accented character splits them:
`"abcédef"` reports `length === 7` and returns a 4-unit `substring(4)` of
its last 3 characters, starting mid-`é`.

`lastIndex` is the load-bearing case. It is both JS-visible and the
engine's own cursor, so it was holding a byte offset in a property the spec
defines in code units — a number the caller cannot use to slice the string
it came from. Walking `/./g` over that string visited `0,1,2,3,5,6,7`,
skipping an index that exists and naming one that does not.

## What changed

Offsets convert at the boundary rather than materializing strings as
`[]uint16`, so the ASCII path stays allocation-free and byte-wise searches
still run through `strings.Index` with their results mapped back before
they reach the caller. An offset landing between the halves of a surrogate
pair rounds up to the end of that character, since a UTF-8 buffer cannot
hold half of one.

- `String.prototype`: `substring`, `slice`, `substr`, `indexOf`,
  `lastIndexOf`, `startsWith`, `endsWith`.
- Match positions, eight sites: both `exec` paths, `Symbol.match`'s fast
  path, `Symbol.replace`'s two fast paths and its spec path, `matchAll`,
  and the two `String.replace` callbacks that pass a position for a string
  search.

Converting `exec()`'s `index` is not sufficient on its own. `Symbol.replace`
reads that property back and uses it to slice `str`, so one number was
serving as both a JS index and a byte offset. They are now separate
variables, `position` for what the callback sees and `positionByte` for
every cut into `str`; their absence is what caused the original bug.

The two halves are one commit each, so they can be reviewed separately.

## Verification

`tests/scripts/string_utf16_indexing.ts` and
`tests/scripts/regex_index_utf16.ts` are new, and cover each fixed method
against a non-ASCII string.

On Test262, `built-ins/RegExp` drops two failures,
`prototype/exec/u-lastindex-value.js` and `S15.10.2.7_A2_T1.js`, with no
regressions. `built-ins/String` is unchanged, with an identical failing
set. `language` shows no attributable change; a batch run of that suite is
not repeatable to better than a few tests here, with and without the patch
alike, so a small number either way is not signal.

Octane's RegExp workload is the thread that led here: `computeInputVariants`
rebuilds a string around one rotated character using `substring`, so every
variant of a non-ASCII input came out two bytes short. This fix accounts
for all of block 6's drift against V8 and all but 1 of block 1's. The
workload's checksum still fails, by less: blocks 0 and 2 have a separate
cause, which is that `\s` and `.` do not carry their ECMAScript meanings
under RE2.

### DIFF
--- a/pkg/builtins/regexp_init.go
+++ b/pkg/builtins/regexp_init.go
@@ -339,17 +339,21 @@ func (r *RegExpInitializer) InitRuntime(ctx *RuntimeContext) error {
 		isSticky := regex.IsSticky()
 
 		if isGlobal || isSticky {
-			// Use lastIndex for stateful matching
+			// lastIndex is JS-visible, so it is held in UTF-16 code units and
+			// converted to a byte offset only to cut the string. Storing the byte
+			// offset instead made `re.lastIndex` report a number the caller could
+			// not use — 6 where the spec says 5 on a string holding one `é`.
 			baseIndex := regex.GetLastIndex()
 			if baseIndex < 0 {
 				baseIndex = 0
 			}
-			if baseIndex > len(str) {
+			if baseIndex > utf16Len(str) {
 				// lastIndex beyond string length - no match possible
 				regex.SetLastIndex(0)
 				return vm.BooleanValue(false), nil
 			}
-			searchStr := str[baseIndex:]
+			baseByte := utf16ToByteOffset(str, baseIndex)
+			searchStr := str[baseByte:]
 			loc := regex.FindStringIndex(searchStr)
 
 			// For sticky flag, match must occur at exactly position 0 of searchStr
@@ -359,7 +363,7 @@ func (r *RegExpInitializer) InitRuntime(ctx *RuntimeContext) error {
 
 			if loc != nil {
 				// Update lastIndex
-				regex.SetLastIndex(baseIndex + loc[1])
+				regex.SetLastIndex(byteToUTF16Offset(str, baseByte+loc[1]))
 				return vm.BooleanValue(true), nil
 			} else {
 				// No match - reset lastIndex
@@ -409,23 +413,26 @@ func (r *RegExpInitializer) InitRuntime(ctx *RuntimeContext) error {
 			if baseIndex < 0 {
 				baseIndex = 0
 			}
-			if baseIndex > len(str) {
+			if baseIndex > utf16Len(str) {
 				// lastIndex beyond string length - no match possible
 				regex.SetLastIndex(0)
 				return vm.Null, nil
 			}
 
+			// lastIndex is in code units; the finder wants a byte offset.
+			baseByte := utf16ToByteOffset(str, baseIndex)
+
 			// Search from baseIndex position (preserves lookbehind context for regexp2)
-			loc = regex.FindStringSubmatchIndexAt(str, baseIndex)
+			loc = regex.FindStringSubmatchIndexAt(str, baseByte)
 
 			// For sticky, match must occur exactly at baseIndex
-			if isSticky && loc != nil && loc[0] != baseIndex {
+			if isSticky && loc != nil && loc[0] != baseByte {
 				loc = nil
 			}
 
 			if loc != nil {
 				// Update lastIndex
-				regex.SetLastIndex(loc[1])
+				regex.SetLastIndex(byteToUTF16Offset(str, loc[1]))
 			} else {
 				// No match - reset lastIndex
 				regex.SetLastIndex(0)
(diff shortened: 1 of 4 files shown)


nooga (approved):
Verified locally: merges cleanly against current main, builds clean, go vet clean, full go test ./... green. Hand-checked the utf16ToByteOffset/byteToUTF16Offset conversion helpers - surrogate-pair rounding is correct, and indexOf/lastIndexOf correctly special-case -1 before converting back to code units. Traced position/positionByte separation through the regexp replace path - consistent throughout. New tests (string_utf16_indexing.ts, regex_index_utf16.ts) pass. Reproduced the Test262 claim directly: S15.10.2.7_A2_T1.js and prototype/exec/u-lastindex-value.js both fail on main, pass here; full built-ins/String and built-ins/RegExp suites diffed against a main baseline show String unchanged (+0/+0) and RegExp +2/-0, matching the PR's numbers exactly.

PR #89, opened 2026-08-29
fix(compiler): route ++/-- on a spilled local through its spill slot

### LINKED ISSUE #27: compiler: fused ++/-- on a spilled local targets R0 instead of the spill slot
## Summary

The fused `++`/`--` path reads a local's `.Register` without checking `IsSpilled`, so an increment/decrement on a **spilled** local targets `R0` instead of the variable's spill slot. The variable never updates (and R0 is corrupted).

## Where

`pkg/compiler/compile_expression.go:705` and `:719` set `identInfo.targetReg = symbolRef.Register` for a local without an `!IsSpilled` guard; that reg is later used by `emitFusedUpdate` (~line 883). A spilled symbol's `Register` defaults to `0` (`!= nilRegister` = 255), so it aliases R0.

## Repro (270 locals force the counter to spill)

```ts
function f(): number {
  let v0 = 0; /* ... let v1 .. v269 ... */ let v269 = 269;
  let c = 100;
  c++;
  return c;   // returns 100 — the ++ hit R0, not c's spill slot
}
```

Actual: `100`. Expected: `101`. (`c--`, `--c`, `++c` behave the same way.)

## Relationship

Surfaced during the audit for #19. Same *class* of defect (code reusing a spilled symbol's zero-valued `Register`), but on the fused-update **use** path, not a binding-**init** path — so it was deliberately left out of #19's scope. Fix: check `IsSpilled` first and route the fused update through the spill slot (emitLoadSpill → update → emitStoreSpill), or fall back to the non-fused path for spilled locals.

mparrett:
Fixes #27.

The fused `++`/`--` path read a local's `Symbol.Register` without checking `IsSpilled`. A spilled symbol leaves `Register` at its zero value, which aliases R0, so the update hit R0 instead of the variable's spill slot.

## Two corrections to the issue as filed

**It is worse than a lost update.** The issue reports `c++` returning the unchanged value, which is what a statement-context `c++` looks like. In an expression context the result is `NaN`:

| | before | after |
|---|---|---|
| `let r = c++; r + "," + c` | `NaN,100` | `100,101` |
| `let r = ++c; r + "," + c` | `NaN,100` | `101,101` |
| `let a = c--; let b = --c` | `NaN,NaN,100` | `100,98,98` |
| `for (…) { c++ }` (×10) | `100` | `110` |

**The threshold is 183 locals, not 270**, in that repro's shape — 182 works, 183 does not. That makes it more reachable than the issue implies, particularly in generated code.

I could not reproduce the "R0 is corrupted" symptom as user-visible damage: with 270 locals, `v0`, `v1`, `v2` and `v269` all kept their values. R0 evidently holds no named local here, so the corruption is real but only observable as the update going missing.

## What changed

Both identifier branches were affected — same-scope (`:705`) and outer-block-scope (`:719`) — so a counter incremented inside a loop body was as broken as a direct `c++`. `c += 1` was always correct, which is what confines this to the fused path.

The current value is loaded from the slot before the fused opcode and stored back after it. The opcode keeps doing the work, including its prefix/postfix result semantics; only where the value lives changes.

`storeToLvalue` had the same defect in its identifier branch — moving into `info.targetReg` with no spill check — and is fixed alongside.

## One thing worth your eyes

`updateIdentInfo` is now a named type. `storeToLvalue` receives it as an `interface{}` and type-asserts it back, and that assertion is **structural**: adding a field to the anonymous struct compiled cleanly and then panicked at run time in a caller that never mentions the new field. `TestCompilationSmokeTests/Update_Expression` caught it, but only because that path happened to be covered. Naming the type makes the two sites share one definition instead of agreeing by coincidence.

`memberInfo` and `indexInfo` are asserted the same way a few lines below and have the same latent fragility. I left them alone to keep this diff to the bug, but they are worth the same treatment.

## Verification

`tests/scripts/update_spilled_local.ts` covers postfix and prefix values, both decrements, and the loop-body shape that reaches the second branch. It was checked against the unfixed compiler and fails there with `NaN,100 | NaN,100 | NaN,NaN,100 | 100`. It uses 270 locals rather than 183 for margin — the test is worthless if the counter stays in a register.

Test262 `language` is unchanged at 23255 passing of 23648, with an identical failing set.


nooga (approved):
Root cause is correct and well-documented: a spilled symbol's Register is 0 (aliases R0), and the fused ++/-- path plus storeToLvalue's identifier branch both dereferenced it without an IsSpilled check. Fix routes through the spill slot in all three sites. Naming updateIdentInfo to kill the structural type-assertion fragility is a good call. Regression test verified to fail pre-fix. Local: build OK, TestScripts green, compiler+vm packages green.

nooga (approved):

Issue #123, opened 2026-08-30
vm: Object.preventExtensions and Object.isExtensible ignore function/RegExp objects; freeze/seal are no-ops on a callable with no properties yet


nooga:
Follow-up to #122. Not a regression — verified identical on a pre-#122-fix binary.

Functions and RegExps keep their ordinary own properties in a side table
(`FunctionObject.Properties`, `ClosureObject.Properties`,
`NativeFunctionObjectWithProps.Properties`, `BoundFunctionObject.Properties`,
`RegExpObject.Properties`). The #122 PR taught `Object.freeze` and
`Object.seal` to reach into that table, which is what `Object.isFrozen` /
`Object.isSealed` already did. Three gaps remain:

## 1. `Object.preventExtensions` is a no-op on callables

```js
var f = function () {};
Object.preventExtensions(f);
f.bar = 2;
console.log(f.bar); // 2 — Node prints undefined (sloppy) / throws (strict)
```

`objectPreventExtensionsWithVM` (`pkg/builtins/object_init.go`) only handles
`TypeArray`, `TypeObject` and `TypeDictObject`.

## 2. `Object.isExtensible` hardcodes `true` for callables

`objectIsExtensibleWithVM` returns `true` for every function type without
consulting the properties table, so it disagrees with `Object.isFrozen`
after `Object.freeze(f)` on a function that has properties, and would
disagree with (1) once that is fixed.

## 3. Integrity levels can't be recorded on a property-less callable

```js
var f = function () {};
Object.freeze(f);
f.x = 1;
console.log(f.x, Object.isFrozen(f)); // 1 false — Node prints undefined true
```

`Object.freeze` is still a no-op here because the properties table is
created lazily and is `nil` until user code writes a property, so there is
nowhere to record the integrity level. Fixing it means allocating the table
on demand, which needs a shared accessor over the five `Properties` fields
above.

Note while doing (3): the lazy creations are inconsistent. Some sites use
`NewObject(Undefined).AsPlainObject()` (extensible: true), others the bare
literal `&PlainObject{prototype: Undefined, shape: RootShape}` — which
leaves `extensible` false. That's inert today because nothing consults it
for callables, but it goes live the moment `isExtensible` starts reading
the table. `grep -rn 'shape: RootShape}' pkg/vm` finds them.

Also missing: `pkg/vm/op_setprop.go`'s `TypeFunction` / `TypeClosure` /
`TypeRegExp` paths call `Properties.SetOwn` directly, which honours
[...]

PR #124, opened 2026-08-30
fix(vm): Object.freeze/seal/defineProperty mutated a shared Shape (#122)

### LINKED ISSUE #122: vm: a property slot that once held a frozen value stays frozen after being reassigned a fresh, unfrozen value
## Summary

A property that once held a value read from a **frozen** object, when later
reassigned (on the same object, same property name) to a brand-new,
unrelated, unfrozen value, keeps behaving as if it's still frozen —
mutating the new value throws `TypeError: Cannot assign to read only
property 'length' of object`, even though the new value was never frozen
itself.

Found running the real, unmodified `pi-coding-agent@0.80.2` npm tree through
`noderati` (highlight.js's `lib/languages/mercury.js`, a real-world instance
of exactly this shape — see "In the wild" below), once #121's fix let the
real `highlight.js` compile at all.

## Repro

```js
const source = Object.freeze({ x: Object.freeze([1]) });
const obj = {};
obj.x = source.x;   // obj.x now references a frozen array, read through a frozen container
obj.x = [9, 9];      // reassign obj.x to a brand-new, plain, unfrozen array literal
obj.x.push(3);       // should succeed — throws instead
console.log(JSON.stringify(obj.x));
```

```
$ paserati -no-typecheck repro.js
PS4001 [ERROR]: Uncaught exception: TypeError: Cannot assign to read only property 'length' of object
    at <script> (<script>:5:1)
  5: obj.x.push(3);
```

Node/V8 print `[9,9,3]`.

## Isolation

Each of these changes to the repro above makes it pass (so all of these
conditions are needed together, not just "a frozen value was involved
somewhere"):

- **Don't freeze `source` itself** (only `x`'s array) — passes. `source`
  being frozen, not just the array it holds, is required.
- **Use a different property name for the reassignment** (`obj.qqq =
  source.zzz; obj.qqq = [9, 9];`) — passes. Same object, same kind of value
  flow, only the name differs.
- **Never actually assign the frozen value into `obj.x`** (read
  `source.x` and discard it, then do a first-ever `obj.x = [9, 9]`) —
  passes. So it's not purely a name-keyed global cache independent of
  value flow either — the frozen value has to actually flow into that
  specific property first.
- **Assign to a brand-new local variable instead of writing back to the
  property** (`const fresh = obj.x; ... ` used directly, no second
  property write) — passes; `Object.isFrozen()` on a fresh local
  correctly reports `false`. The staleness is specific to the *property
  slot*, not to the array values in play.

So the trigger is exactly: property `P` on object `O` is set to a value
that was read from a frozen object, then `P` on that same `O` is
overwritten again — and something about that first write leaves a stale
"frozen" marker attached to the `(O, P)` slot (or its shape) that a later,
unrelated value doesn't clear.

## Where to look

Haven't traced this one to a specific line (unlike #121) — flagging with a
precise, minimal, deterministic repro instead. Likely candidates given the
shape (a stale per-shape/per-slot cache surviving a value overwrite):

- `pkg/vm/op_setprop.go` — the property-assignment opcode; look at whatever
  fast path exists for "property already exists on this shape."
- `pkg/vm/ic_site_cache.go` / `pkg/vm/cache.go` — the inline cache; look for
  anything that memoizes a "writable"/"frozen" bit per (shape, property
  name) at *write* time without being invalidated by the write's own new
  value.
- `pkg/vm/value.go:197` (`ArrayObject.frozen bool`) and `:2486`
  (`SetFrozen`) — confirmed via a separate check that `.slice()`'s actual
  result array is genuinely unfrozen at construction (`vm.NewArrayWithArgs`
  in `pkg/builtins/array_init.go` always builds a fresh array), so the bug
  isn't in array construction — it's specifically that reading the array
  back out through `obj.x` after the second write is what reports (and
  enforces) stale frozen-ness.

## In the wild

`highlight.js`'s `lib/languages/mercury.js` (bundled with
`@earendil-works/pi-coding-agent@0.80.2`) hits this exact shape:

```js
const STRING = hljs.inherit(hljs.QUOTE_STRING_MODE, { relevance: 0 });
// hljs.QUOTE_STRING_MODE is deep-frozen by highlight.js's own core.js at
// module-load time (a bundled deep-freeze-es6, used deliberately so
// language plugins can't accidentally mutate the shared default modes).
// hljs.inherit does `result[key] = original[key]` for each key, so
// STRING.contains now holds the same frozen array as
// hljs.QUOTE_STRING_MODE.contains.
STRING.contains = STRING.contains.slice(); // fresh, unfrozen array by the time it returns
STRING.contains.push(STRING_FMT);          // throws here — same STRING.contains slot
```

`highlight.js` catches the per-language registration error and continues
(it doesn't crash the whole library), but the language never registers, and
noderati/pi print a `Language definition for 'mercury' could not be
registered` error on every run that real Node doesn't.

nooga:
Fixes #122.

## Root cause

Shapes are shared. Every object that adds the same properties in the same order
lands on the same `*Shape` through the transition maps — that's the whole point
of them. But `FreezeAllProperties`, `SealAllProperties` and the four
`DefineOwnProperty`/`DefineAccessorProperty` update paths all mutated
`o.shape.fields[i]` **in place**. `privatizeFields()` privatized the fields
*backing array*, not the Shape identity, so the attribute change was visible to
every object sharing that shape — before, after, and entirely unrelated:

```js
const a = { x: 1 }, b = { x: 1 };
Object.freeze(a);
b.x = 42;             // silently ignored — b's "x" is read-only too
console.log(b.x);     // 1
```

```js
const hidden = { k: 1 }, visible = { k: 1 };
Object.defineProperty(hidden, "k", { enumerable: false });
Object.keys(visible); // [] — should be ["k"]
```

That's what #122's repro hits: `obj.x = source.x` transitions `obj` onto the
shape `Object.freeze(source)` poisoned, so the *next* write to `obj.x` is
rejected and `obj.x` keeps the old frozen array.

## Fix

Attribute changes fork instead of mutating:

- **`forkShape()`** gives the object a private, unshared Shape (kept out of any
  transition map) before the in-place field write — the same thing
  `DeleteOwnByKey` already did for property removal.
- **`attrTransition()`** handles freeze/seal, forking with a memo on the source
  shape, so objects that shared a shape before `Object.freeze` still share one
  frozen shape afterwards and inline caches stay monomorphic.
- `ownedFields` is removed: it existed only to serve `privatizeFields`, and
  documented an invariant ("mutate `fields` in place after privatizing") that
  licensed the bug in the first place.

Inline caches guard on shape pointer *and* version, so a fresh shape is a clean
miss; the IC is the only consumer of `Shape.version`.

## Two more bugs of the same family, found on the way

- `DefineAccessorPropertyByKey` cached its new-field transition under
  `key.hash()` alone, so a second `defineProperty` for the same key with
  different `enumerable`/`configurable` silently inherited the first one's
  attributes. The cache key now folds the attributes in.
- Built-in constructors were getting `prototype` as `{writable: false,
  enumerable: false, configurable: false}` **by accident**: the first
  constructor to lock its prototype down re-attributed every other one's
  through the shared shape. Once shapes fork, each site has to say what it
  means, so the ~29 sites that only called `SetOwnNonEnumerable` now use
  `DefineFixedProperty`. (Without this the diff was -15; this is what took it
  back to positive.)

## Also: freeze/seal on functions and RegExps

`Object.freeze`/`Object.seal` were no-ops on callables and RegExps — their own
properties live in a side table that neither built-in looked at, even though
`Object.isFrozen`/`isSealed` already read it, so the two disagreed. Both now
reach the table.

This is also what fixed the last two `-` lines in the built-ins diff
(`Object/freeze/15.2.3.9-2-a-9.js` and `Object/seal/...function-object...`),
which had been passing only because the shared-shape leak poisoned the shape
they check — see #123 for why that counts as a fix rather than a restoration.

The remaining gaps there (`preventExtensions`/`isExtensible` on callables,
freezing a callable that has no properties yet) are pre-existing and filed
separately as #123 — confirmed against a pre-change build, not caused by this
PR.

## Verification

- `tests/scripts/frozen_shape_isolation.ts` covers the original repro plus the
  simpler root-cause shapes (sibling writability, `Object.keys` leak,
  seal-then-freeze, double freeze, accessor attributes, constructor
  `prototype`).
- `go test ./tests -run TestScripts` green; full `go test ./tests/... ./pkg/...`
  green; `go test -race ./pkg/vm/... ./tests/...` green.
- test262 `built-ins`: **+2 / -0** against `baseline.txt`.
- test262 `language`: **+1 / -0** against a dump from a pre-change binary
  (`baseline.txt` is built-ins only). `annexB`: +0 / -0.
- `cmd/bench-ratchet`: 0 regressions over budget (the tree at `main` reports 1
  on this machine).

### DIFF
--- a/pkg/builtins/abort_controller_init.go
+++ b/pkg/builtins/abort_controller_init.go
@@ -120,7 +120,7 @@ func (a *AbortControllerInitializer) InitRuntime(ctx *RuntimeContext) error {
 		return createAbortSignalObject(vmInstance, signal, signalProto), nil
 	}))
 
-	signalConstructor.SetOwnNonEnumerable("prototype", vm.NewValueFromPlainObject(signalProto))
+	signalConstructor.DefineFixedProperty("prototype", vm.NewValueFromPlainObject(signalProto))
 
 	if err := ctx.DefineGlobal("AbortSignal", vm.NewValueFromPlainObject(signalConstructor)); err != nil {
 		return err
@@ -145,7 +145,7 @@ func (a *AbortControllerInitializer) InitRuntime(ctx *RuntimeContext) error {
 	controllerConstructor := vm.NewConstructorWithProps(0, false, "AbortController", controllerConstructorFn)
 	if controllerConstructor.Type() == vm.TypeNativeFunctionWithProps {
 		ctorProps := controllerConstructor.AsNativeFunctionWithProps()
-		ctorProps.Properties.SetOwnNonEnumerable("prototype", vm.NewValueFromPlainObject(controllerProto))
+		ctorProps.Properties.DefineFixedProperty("prototype", vm.NewValueFromPlainObject(controllerProto))
 	}
 
 	controllerProto.SetOwnNonEnumerable("constructor", controllerConstructor)
--- a/pkg/builtins/array_init.go
+++ b/pkg/builtins/array_init.go
@@ -2280,7 +2280,7 @@ func (a *ArrayInitializer) InitRuntime(ctx *RuntimeContext) error {
 	})
 
 	// Add prototype property
-	ctorWithProps.AsNativeFunctionWithProps().Properties.SetOwnNonEnumerable("prototype", vm.NewValueFromPlainObject(arrayProto))
+	ctorWithProps.AsNativeFunctionWithProps().Properties.DefineFixedProperty("prototype", vm.NewValueFromPlainObject(arrayProto))
 
 	// Add static methods
 	ctorWithProps.AsNativeFunctionWithProps().Properties.SetOwnNonEnumerable("isArray", vm.NewNativeFunction(1, false, "isArray", func(args []vm.Value) (vm.Value, error) {
--- a/pkg/builtins/arraybuffer_init.go
+++ b/pkg/builtins/arraybuffer_init.go
@@ -369,7 +369,7 @@ func (a *ArrayBufferInitializer) InitRuntime(ctx *RuntimeContext) error {
 	})
 
 	// Add prototype property
-	ctorWithProps.AsNativeFunctionWithProps().Properties.SetOwnNonEnumerable("prototype", vm.NewValueFromPlainObject(arrayBufferProto))
+	ctorWithProps.AsNativeFunctionWithProps().Properties.DefineFixedProperty("prototype", vm.NewValueFromPlainObject(arrayBufferProto))
 
 	// Add static methods
 	ctorWithProps.AsNativeFunctionWithProps().Properties.SetOwnNonEnumerable("isView", vm.NewNativeFunction(1, false, "isView", func(args []vm.Value) (vm.Value, error) {
--- a/pkg/builtins/bigint_init.go
+++ b/pkg/builtins/bigint_init.go
@@ -262,7 +262,7 @@ func (b *BigIntInitializer) InitRuntime(ctx *RuntimeContext) error {
 		return vm.NewBigInt(result), nil
 	}))
 
-	bigintConstructor.AsNativeFunctionWithProps().Properties.SetOwnNonEnumerable("prototype", vmInstance.BigIntPrototype)
+	bigintConstructor.AsNativeFunctionWithProps().Properties.DefineFixedProperty("prototype", vmInstance.BigIntPrototype)
 
 	// Set constructor property on prototype
 	bigintProto.SetOwnNonEnumerable("constructor", bigintConstructor)
--- a/pkg/builtins/blob_init.go
+++ b/pkg/builtins/blob_init.go
@@ -93,7 +93,7 @@ func (b *BlobInitializer) InitRuntime(ctx *RuntimeContext) error {
 	blobConstructor := vm.NewConstructorWithProps(2, false, "Blob", blobConstructorFn)
 	if blobConstructor.Type() == vm.TypeNativeFunctionWithProps {
 		ctorProps := blobConstructor.AsNativeFunctionWithProps()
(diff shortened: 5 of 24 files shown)


mparrett:
Late measurement on this one, since it is easier to act on here than in a
standalone issue. Filed as #150 with the full detail.

**This costs about 7% on prototype-chain lookup.** `a1e5e22e` against its
immediate parent `acd1d7fa`, one dedicated `c7a.2xlarge`, 3 rounds interleaved,
anchor drift 0.00%:

| benchmark | median Δ | per round |
|---|---|---|
| `PrototypeMethodAccess/WithPrototypeCache/ObjectPrototypeChain` | **+7.8%** | +6.5, +7.8, +8.5 |
| `PrototypeMethodAccess/Baseline/ObjectPrototypeChain` | **+7.0%** | +5.3, +7.3, +7.0 |
| `PrototypeMethodAccess/WithDetailedStats/ObjectPrototypeChain` | **+6.8%** | +6.8, +5.0, +8.8 |

Median across all 46 benchmarks is +0.00%, so it is targeted at prototype-chain
walking rather than a general slowdown, and it persists to current `main`.

**How much is real:** the same session measured a layout null control — a commit
adding 27 lines to `pkg/vm` that cannot change behaviour — and it moved 10 of 42
benchmarks by >2%, max 4.2%, with this family the most layout-sensitive in the
corpus. So ~7% clears that floor and the effect is real, but its size is
uncertain; I would not defend it more precisely than mid-single to low-double
digits.

Nothing here argues against the fix — the shared-Shape bug was real and worth
fixing. This is the price tag, in case it is worth a second look at the read path
`forkShape()` added, or in case it is simply the cost of correctness and you would
rather pay it. Your call; I have no PR to offer.

Issue #142, opened 2026-08-31
vm: an ordinary throw across a *reentrant* vm.Call() (e.g. require() nested inside require()) still loses the exception - caught value is null


nooga:
## Summary

`#130`'s fix (fa0d5451) correctly surfaces the VM's own recorded diagnostic
for the case it targeted: an internal-invariant failure through
`vm.runtimeError()` (corrupted bytecode, stack overflow, etc.) with no real
JS exception value to report. Confirmed working via the issue's own added
tests.

But the *original* #130 repro — an ordinary JS `throw` (a real exception,
not an internal invariant failure) propagating out through a `vm.Call()`
that's itself invoked from a native Go function called from bytecode
already running under another `vm.Call()` — still loses the real exception.
Uncaught, it still reports the same generic
`runtime error during user function execution`. Caught, it's worse: the
value the `catch` block receives is literally `null`, not even a
generic/wrong `Error` object.

This looks like a different mechanism than what #130's fix addressed (which
explicitly did not find/need a reentrancy-based root cause — see its commit
message: "explicitly did NOT find the reentrancy-specific state corruption
#130 hypothesized... The actual trigger needs no second
executeUserFunctionSafe on the Go stack at all"). This report is about the
case where there *is* a second `executeUserFunctionSafe`/`vm.Call` on the Go
stack, triggered by an ordinary throw, not an internal invariant failure —
apparently not what the #130 investigation tested.

## Repro

Two tiny files, no npm packages (same as originally reported on #130):

```js
// b.js
throw new Error("boom from b.js");
```
```js
// a.js  (noderati's CJS require(), which is exactly two Go-level vm.Call
// invocations: one for a.js itself, a second — from inside a native
// require() function called from a.js's own running bytecode — for b.js)
require("./b.js");
console.log("unreachable");
```

```
$ noderati a.js
PS4001 [ERROR]: runtime error during user function execution
```

**Caught**, the value is `null`, not an `Error`:

```js
try {
  require("./b.js");
} catch (err) {
  console.log(typeof err, err === null, JSON.stringify(err));
}
```
```
object true null
```

Node prints the real `Error` and, in the caught version, a real `Error`
instance with `.message === "boom from b.js"`.

Confirmed via `#130`'s own now-passing tests that the *runtimeError()* path
is genuinely fixed — this is additional, not a regression: a stack-overflow-
style internal failure through a single `vm.Call()` surfaces correctly now;
[...]

Issue #130, opened 2026-08-30
vm: a reentrant vm.Call() (e.g. require() of a module that throws) loses the real exception, reporting a generic message instead


nooga:
## Summary

`vm.Call()` invoked on a user function while *another* `vm.Call()` /
`executeUserFunctionSafe()` invocation is already active further up the Go
call stack loses the real thrown exception if the inner call's body throws
synchronously: instead of surfacing the real `Error`, the caller gets a
generic, useless `runtime error during user function execution` with no
message, type, or location. A single, non-reentrant `vm.Call()` that throws
surfaces the real exception correctly — this is specific to the nested case.

Found running the real, unmodified `pi-coding-agent@0.80.2` npm tree
through `noderati`: a CJS module's `require()` of a module whose top-level
code throws (e.g. `graceful-fs`'s `polyfills.js` doing `require('constants')`
for noderati's still-missing `constants` builtin — a real, separate,
noderati-side gap, not a paserati issue on its own) reports nothing useful,
making it impossible to tell what actually went wrong without bisecting the
require graph by hand.

## Repro

Two tiny files (no npm packages involved):

```js
// b.js
throw new Error("boom from b.js");
```
```js
// a.js
require("./b.js");
console.log("unreachable");
```

```
$ noderati a.js
PS4001 [ERROR]: runtime error during user function execution
```

Node prints the real error with a stack trace pointing at `b.js`.

A single-level throw (the *entry* file itself throwing, only one `vm.Call`
in play) reports correctly:

```js
throw new Error("boom directly at top level");
```
```
PS4001 [ERROR]: VM exception: {..., name: "Error", message: "boom directly at top level", ...}
```

And a native→JS callback that isn't nested inside another `vm.Call` also
reports correctly, e.g. `[1].forEach(() => { throw new Error("boom") })`
surfaces the real `Error` with a full stack. So this isn't "callbacks lose
their exception" in general — it's specific to a `vm.Call` that's already
running *inside* another native call that itself reached the VM via
`vm.Call`/`executeUserFunctionSafe`. Confirmed the depth matters, not just
"require is special": a 3-level require chain (`a2.js` → `b2.js` → `c.js` →
`d.js` throws) hits the same generic message as the 2-level case.

## Root cause (traced through noderati's CJS loader into the VM)

noderati's CJS `require()` (`internal/host/cjs.go`, `execFile`) invokes each
required module's body via `vmInst.Call(fn, ...)`
[...]

Issue #128, opened 2026-08-30
compiler: a function-scoped class declaration isn't visible to closures inside that function (upvalue capture falls through to a bogus global)


nooga:
## Summary

A `class` declared inside a function body cannot be referenced from *any*
closure nested inside that same function — including, but not limited to,
the class's own methods. The reference resolves as a (non-existent) global
instead of an upvalue capture, throwing `ReferenceError: X is not defined`
at call time. `let`/`const`/plain function declarations in the identical
position work correctly; only `class` is affected.

This is the same failure family as #117/#119 (a not-yet-finalized top-level
binding falling through to "must be a global"), but for local (function- or
block-scoped) class declarations captured by upvalue, rather than
module-level declarations read via `OpGetGlobal`.

Found running the real, unmodified `pi-coding-agent@0.80.2` npm tree through
`noderati`: `hosted-git-info`'s `GitHost` class (CJS, so noderati runs it
through its function-wrapped CJS interop) breaks this way on its own static
methods (`GitHost.addHost`, `GitHost.fromUrl`, etc., all of which reference
`GitHost` by name).

## Repro

Minimal, no self-reference needed — a plain, unrelated sibling closure
capturing an outer function's class is enough:

```js
function make() {
  class Foo {}
  function inner() { return Foo; }
  return inner;
}
console.log(typeof make()());
```

```
$ paserati -no-typecheck repro.js
PS4001 [ERROR]: Uncaught exception: ReferenceError: Foo is not defined
    at inner (<inner>:3:1)
```

Node prints `function`.

The originally-found shape (a class's own method self-referencing the
class, also inside a function):

```js
function make() {
  class Foo {
    static bar() { return Foo; }
  }
  return Foo;
}
console.log(make().bar() === make());
```

Same error, same location shape (`ReferenceError: Foo is not defined` inside
`bar`).

## Isolation

- **Top-level (module or script) self-referencing class** — works. `class
  Foo { static bar() { return Foo; } }` at the top of a file, whether run as
  a module (`export class Foo ...`) or a plain top-level script, resolves
  `Foo` correctly (this is #119's fix — global-indexed, not register/upvalue
  based).
- **Block-scoped (not function-scoped) self-referencing class at top level**
  — works: `{ class Foo { static bar() { return Foo; } } console.log(Foo.bar() === Foo); }`
  at the top level prints `true`.
- **Plain closure over an outer function's `const`** — works:
  `function make() { const x = 42; function inner() { return x; } return inner; }`
[...]

Issue #65, opened 2026-08-26
test262 runner (and vm.Interpret) score a script as passing when execution silently aborts at a native exception boundary


nooga:
## Summary

`paserati-test262` (and the underlying `driver.Interpret`/`InterpretChunk` API it uses) can score a test as **PASS** when the script's execution actually aborted partway through with an unrecoverable exception, because the failure never gets recorded into `vm.errors`.

This means the project's headline Test262 compliance numbers are inflated by an unknown amount - some fraction of tests currently counted as passing never actually complete.

## Mechanism

`cmd/paserati-test262/main.go` decides pass/fail purely on `len(runtimeErrs) == 0` from `paserati.InterpretChunk(chunk)` (main.go around line ~770, mirrored in the module-mode path around line ~660-700):

```go
_, runtimeErrs := paserati.InterpretChunk(chunk)
if len(runtimeErrs) > 0 {
    // ... fail ...
}
// Success
```

`InterpretChunk` → `vm.Interpret(chunk)` (pkg/vm/vm.go:1112) returns `vm.errors` verbatim:

```go
if resultStatus == InterpretRuntimeError {
    return finalValue, vm.errors
}
```

`vm.errors` is only populated by `handleUncaughtException()` (pkg/vm/exceptions.go:365), which `throwException` only calls when `unwindException()` reports **no handler found** (exceptions.go:118-141). But `unwindException` also returns "handled" (true) when it stops at a native/direct-call boundary via `vm.unwindingCrossedNative` (pkg/vm/exceptions.go, and the ~300 `if vm.frameCount == 0 || vm.unwindingCrossedNative { ... }` dispatch-loop checkpoints throughout pkg/vm/vm.go, e.g. line 1339) - that path assumes a native Go caller further up the stack (`vm.Call`, `executeUserFunctionSafe`, etc.) will receive the exception as a Go `error` and either re-throw it or handle it.

[...]

PR #143, opened 2026-08-31
fix(compiler,vm): hoisted/earlier-closure class forward refs (#141, #144) and reentrant vm.Call() exception leak (#142)

### LINKED ISSUE #141: compiler: #128's fix doesn't cover a hoisted function declaration referencing a sibling function-scoped class
## Summary

`#128`'s fix (357881e2) correctly makes a function-scoped `class` visible to
arrow functions, object/class methods, and immediately-nested closures — but
a **hoisted function declaration** referencing the same class, as a sibling
in the same enclosing function, still isn't visible to it. Same
`ReferenceError: X is not defined` symptom, same `OpGetGlobal`-instead-of-
upvalue bytecode shape as the original #128, just for one specific closure
kind the fix didn't cover.

## Repro

```js
function make() {
  class Foo {}
  function inner() { return Foo; }   // hoisted function *declaration*
  return inner;
}
console.log(typeof make()());
```

```
$ paserati -no-typecheck repro.js
PS4001 [ERROR]: Uncaught exception: ReferenceError: Foo is not defined
    at inner (<inner>:3:1)
```

Node prints `function`.

## Isolation — confirms this is specifically about hoisting, not escape or invocation timing

- **`#128`'s own regression test** (`tests/scripts/class_visible_to_closures_in_function.ts`)
  covers "a sibling closure declared after the class in the same function
  scope" via `const siblingReadsClass = () => Counter.count;` — an arrow
  function, assigned to a `const`. That passes (confirmed by re-running the
  test after pulling the fix). The *only* structural difference in this
  repro is `function inner() { ... }` (a declaration) instead of
  `const inner = () => { ... }` (an arrow function expression assigned to a
  binding) — same position, same body, same call pattern.
- **Not about escaping the enclosing function**: an arrow function that
  *also* escapes (returned and called after `make()` has already returned,
  identical to this repro's shape) works fine —
  `const inner = () => Foo; return inner;` then `make()()` prints
  `function` correctly.
- **Not about invocation timing**: a function *declaration* called
  **immediately**, still inside `make()`'s own execution (no escape at all)
  — `function inner() { return Foo; } return inner();` — still throws the
  same error. So this isn't about the closure outliving its enclosing
  frame; a hoisted function declaration referencing a sibling local class
  fails even when called synchronously, in place.

## Root cause (traced via `-bytecode`, same technique as the original #128 report)

`inner`'s compiled body is still:

```
== inner ==
0000      OpGetGlobal      R1, GlobalIdx 80
0004      OpReturn         R1
```

— a brand-new, never-written global index, not an upvalue capture of
`Foo`'s real local slot — the exact same symptom #128 fixed for other
closure kinds. Presumably: function declarations are hoisted and
processed in an earlier pass than the sequential statement compilation
`compileClassDeclaration` participates in, so `inner`'s free-variable
resolution for `Foo` runs before the class's now-stable spill slot
(357881e2's fix) has been set up — mirroring the exact `#117`-vs-`#119`
split from the original module-level version of this bug family: `#117`
covered general hoisted-function access to a sibling top-level
`let`/`const`, but hoisted-function access to a sibling `var`/`class`
specifically needed its own follow-up (`#119`) because hoisting order
interacts with declaration-kind differently. This looks like the same
split reappearing one level down, now for function-local scopes.

## Where to look

- `pkg/compiler/compile_class.go` — 357881e2's spill-slot allocation for
  the local-class path (the fix for the original #128). Worth checking
  when it runs relative to the enclosing function's own hoisting pass for
  nested function *declarations* (as opposed to when arrow
  functions/object literals holding closures get compiled, which is purely
  sequential/inline and unaffected by hoisting order).
- Whatever code hoists local function declarations within a function body
  (compiles their signatures/reserves their binding before the function
  body's sequential statements run) — likely the same kind of "hoisted
  function compiled before an as-yet-unprocessed sibling declaration"
  ordering gap `#117`/`#119` fixed for module scope, now needed for
  function-local class declarations specifically.

Found and isolated 2026-08-30 re-verifying `#128`'s fix against the
original `hosted-git-info` motivation (noderati's Phase 3) — `GitHost`'s
own static methods are fixed now, but this is a nearby, still-open gap in
the same area. Not yet attempted to fix.

### LINKED ISSUE #144: compiler: a closure defined before a function-scoped class (not hoisted, just earlier in the block) still can't capture it, even when only called after
## Summary

A closure defined **textually before** a function-scoped `class` in the
same block, but only **called after** that class has been declared (so
there's no real TDZ violation — completely ordinary, common JS), still
can't reference the class: `ReferenceError: X is not defined`, same
`OpGetGlobal`-instead-of-upvalue shape as `#128`/`#141`.

This is #128/#141's underlying issue reappearing a third way. #128 fixed
closures compiled *inside* the class body (constructor/methods/nested
arrows) by pre-allocating the class's spill slot before those compile.
#141 fixed *hoisted function declarations* referencing a later sibling
class, by pre-registering the class's spill slot before a block's hoisted
functions compile. Neither covers an ordinary, non-hoisted closure
(`const f = () => ...`, a plain arrow function or function expression
assigned to a binding) that's simply positioned earlier in the same
block's statement list than the class it references — the compiler
appears to resolve free variables at each statement's own sequential
compile position, and a class's spill slot doesn't exist yet if the
compiler hasn't reached the class's own declaration statement.

Found in a real, extremely common npm package: `minimatch`'s own
`dist/commonjs/index.js` defines `const minimatch = (p, pattern, options)
=> { ...; return new Minimatch(pattern, options).match(p); }` near the top
of the file, with `class Minimatch { ... }` declared later in the same
(CJS-module-wrapped) scope — real Node handles this trivially, since
`minimatch(...)` is never actually called until well after the whole
module has finished loading.

## Repro

```js
function make() {
  const useIt = (x) => new Foo(x).value;   // arrow fn, defined BEFORE the class
  class Foo {
    constructor(x) { this.value = x; }
  }
  return useIt;                             // only called after returning
}
console.log(make()(42));
```

```
$ paserati -no-typecheck repro.js
PS4001 [ERROR]: Uncaught exception: ReferenceError: Foo is not defined
    at useIt (<useIt>:3:1)
```

Node prints `42`. Confirmed against the branch with both `#128` and
`#141`'s fixes applied (`fix/141-142-class-hoisting-and-reentrant-exceptions`,
`3b1fa14e`) — this is not a regression from either, it's a third,
still-open case in the same family.

## Isolation

- **Same shape, arrow function defined *after* the class** (like #128's own
  regression test's "sibling closure" case) — works. Only the *ordering*
  (closure's own source position relative to the class) matters, not
  whether it's an arrow function vs. a hoisted function declaration
  (already covered by `#141`) vs. a method/constructor (already covered by
  `#128`).
- **Confirmed via `-bytecode`**: `useIt`'s body compiles to `OpGetGlobal`
  for `Foo`, not an upvalue capture — identical symptom to #128/#141's own
  bytecode evidence.

## Root cause (not traced to a specific line this time — pattern-matching against #128/#141's own fixes)

Both `#128` and `#141`'s fixes work by pre-registering the class's spill
slot **at a specific point**: before the class's own constructor/methods
compile (#128), or before a block's *hoisted* functions compile (#141).
Neither covers a plain, non-hoisted statement (`const x = () => ...`,
compiled at its own sequential position in the block) that appears earlier
in source order than the class. The compiler's free-variable resolution
for such a closure presumably runs at that closure's own compile time, when
the class hasn't been reached yet and has no storage location at all — not
even the `nilRegister` placeholder #128 originally found, since the class's
declaration-time pre-registration (both before and after #128/#141) only
happens when the compiler's sequential walk reaches the class's own
statement, not earlier.

A general fix likely needs to pre-register spill slots for **every**
class (and probably `let`/`const`/`var`) declared anywhere in a block's own
statement list, before compiling *any* statement in that block — the same
principle #141 applied to hoisted functions specifically, generalized to
apply before sequential compilation of the block begins at all. This would
match how real JS scoping actually works: a block-scoped binding exists
(in TDZ) for the entire block from the start, not just from its own
declaration point onward.

## Where to look

- `pkg/compiler/compile_class.go` — wherever `#128`/`#141`'s spill-slot
  pre-registration was added; likely needs to move earlier, to run before
  a block's statements compile in general, not just before its hoisted
  functions or a class's own body.
- Whatever function compiles a block's statement list sequentially (where
  `#141`'s "pre-register spill slots for classes... before that block's
  hoisted functions compile" logic lives) is probably the right place to
  generalize from "before hoisted functions" to "before any statement in
  the block."

Found and isolated 2026-08-31 verifying PR #143 against noderati's Phase 3
(deleting third-party npm-package fakes) — `minimatch`'s fake looked safe
to delete per an automated check (`pi --version`/`--help`/`-p` never
actually call `minimatch()`), but calling it directly, as real usage would,
hits this immediately. Not yet attempted to fix.

nooga:
## Summary

Three narrower follow-ups in the same family, each filed as its own issue:

- **#141** — a hoisted function declaration still couldn't forward-reference a sibling function-scoped class (#128 covered closures nested *inside* the class body, but not a hoisted function compiled ahead of the class's own position).
- **#142** — an ordinary throw could still get lost across a reentrant `vm.Call()` (two-plus nested `vm.Call()` invocations on the Go stack), independent of #130.
- **#144** — a *non-hoisted* closure (`const f = () => ...`) positioned earlier in a block than a sibling function-scoped class it references still couldn't capture it, even when only called after the class is declared. Found for real in `minimatch`'s CJS bundle while verifying this PR.

Three separate commits, one per issue.

## #141 — hoisted function declarations vs. sibling function-scoped classes

`#128` fixed forward references to a function-scoped class from closures compiled *inside* the class body by pre-allocating a stable spill slot before the constructor/methods compile. It didn't cover a hoisted *function declaration* sibling in the same block — those compile at the very top of the block, before the class's own sequential position, so the reference resolved before the class had anywhere to live.

Fixed by pre-registering spill slots for classes declared directly in a block's own statement list before that block's hoisted functions compile; `compileClassDeclaration` reuses that slot instead of allocating a second one.

**Verified:** `go test ./...` clean, full smoke suite green, Test262 `language/**` diff vs. real `main`: 0 regressions.

## #142 — reentrant `vm.Call()` losing exceptions

`vm.go:11288` (flagged in the issue as worth checking first) turned out to be a dead end — confirmed and documented in the commit.

Root cause: `executeUserFunctionSafe`/`executeUserFunctionWithNewTarget` deliberately left `vm.unwinding = true` after handing an exception off as a Go error (a prior "WIP unwinding fixes" commit, reasoning that downstream code would either make a new `vm.Call` or re-throw). That misses a third case: native code that fully *absorbs* the exception (e.g. turning a `.then()` handler's throw into a rejected promise) and returns normally. Since `vm.unwinding` is checked bare after nearly every VM instruction across `vm.go`, a leaked flag poisons the next instruction *any* `vm.run()` executes, anywhere — reproduced crashing an unrelated top-level `await` with a bogus `Uncaught exception: null`.

Fixed at the source: both functions now clear `vm.unwinding` alongside `vm.currentException` on handoff. Also added a defensive `vm.ClearUnwindingState()` at the actual leak site in `triggerPromiseReactions`, mirroring the existing call in `NewPromiseFromExecutor`.

**Verified:**
- `go test ./...` clean, full smoke suite green.
- New test reproduces the crash when **both** halves of the fix are reverted. Correcting an earlier, overstated claim in this PR: reverting only the `vm_init.go` clear leaves the test green, because `triggerPromiseReactions`' own `ClearUnwindingState()` absorbs the leak one level up. The two are each independently sufficient for that scenario; the test header now carries the measured breakdown.
- Test262 diff vs. real `main`: `built-ins/**` **+21 new passes, 0 regressions** (Promise.all ordering, Array.from iterator/mapFn-throw handling, Iterator.zip suspended-close semantics, Reflect apply/construct argument checks, RegExp/String matchAll-throws propagation, TypedArray Get-trap bounds checks) — that +21 is what the general `vm_init.go` clear buys, and reverting those four lines gives all of it back.
- `language/**` **0 new passes, 0 new failures** — but read that 0 carefully, it is not "clean". Also correcting an earlier "0 regressions" claim here, which was measured against a baseline that already contained this PR. Diffed against real `main`, this PR started out with **two** `language` failures:
  - `with/unscopables-prop-get-err.js` — a **genuine regression**, now fixed. `isUnscopable()` documented its contract as "if hadError, check `vm.unwinding`", which only held while `vm.Call` left the flag set; its `vm.GetProperty` error path returned a bare `hadError=true` without putting the exception back in flight, so a throwing `Symbol.unscopables` getter was silently swallowed. It now re-throws, mirroring the Proxy get-trap branch in the same function. (Deliberately *not* applied to the two `GetSymbolPropertyWithGetter` paths there — that function already throws internally, and re-throwing double-throws and breaks `with/unscopables-get-err.js`.)
  - `types/reference/put-value-prop-base-primitive-realm.js` — **a false pass on both sides; it carries no signal either way.** On `main` the leaked flag aborts that script at its cross-realm `other.eval(...)` before the first assertion runs, and the runner scores a silently aborted script as a pass (#65). Closing the leak let it reach the assertion and fail honestly. Chasing *that* failure turned up a third real bug, now also fixed here: `opSetProp`'s primitive-base path silently discarded exceptions thrown by an accessor setter or a Proxy set trap reached through a primitive's prototype chain (`try { (0).x = 1 } catch {}` never caught, while the identical shape on an object base threw correctly). With the exception now propagating, it aborts the script at the same `other.eval(...)` — so the test is back to "passing" by abort, verified with `print()` markers that it still never reaches its first assertion. **The real coverage for that fix is the new `primitive_base_setter_throw_propagates.ts` smoke test**, which is verified to fail with the fix reverted — not this test262 line.

## #144 — earlier closure vs. sibling function-scoped class

Generalizes #141 rather than being independent: #141 only pre-registered a block's class spill slots before that block's *hoisted functions* compiled. A plain, non-hoisted closure (`const f = () => ...`) compiles at its own sequential position — if that position is textually before the class it references, the same dangling-capture bug hits it too, even though it's only ever *called* well after the class has been declared (no real TDZ violation).

Found for real in `minimatch`'s `dist/commonjs/index.js`: `const minimatch = (p, pattern, options) => { ...; return new Minimatch(...).match(p); }` defined near the top of the file, with `class Minimatch { ... }` declared later in the same (CJS-module-wrapped) scope.

Fixed by widening #141's pre-registration from "before this block's hoisted functions compile" to "before any statement in this block compiles" — same gate (only for classes taking the local/spilled path), just no longer conditioned on the block having hoisted declarations at all.

Checked explicitly whether the true top-level (module/script scope) case of the minimatch example is still broken by the remaining gate: it isn't — a top-level class already takes the global path, where the existing bare-key/`moduleGlobalKey` fallback in `compileIdentifier` already resolves it (same mechanism #141's own commit documented for a bare top-level block). Confirmed by hand.

**Verified:** `go test ./...` clean, full smoke suite green, Test262 diff vs. real `main`: 0 regressions on `language/**` and `built-ins/**`.

## Known, documented, not-fixed-here gaps (shared by all three)

- Type checking still rejects these patterns with `TS2304` — the fixes only apply under `--no-typecheck`/noderati execution. Real, separate TS-conformance bugs.
- A closure (hoisted or not) inside a *nested* block referencing a class declared in an *enclosing* block still throws `ReferenceError` — same shape, deeper cross-scope problem `collectClassDeclarations`'s single-block scan doesn't reach.

- A cross-realm call does not switch to the callee's `[[Realm]]`, so a main-realm Proxy set trap throws `ReferenceError` on its own global when invoked from another realm. Confirmed pre-existing on an `a88125c2` worktree.
- An exception escaping `realmGlobal.eval(...)` silently aborts the script instead of being catchable — `try { other.eval('throw new Error("x")') } catch (e) {}` aborts uncatchably on plain `main` too. #65 territory. These two together are what keep `put-value-prop-base-primitive-realm.js` a false pass.
- The general `vm.unwinding` clear has no *discriminating* smoke test — its coverage is the 21 Test262 built-ins tests. A second absorber that doesn't route through `triggerPromiseReactions` would be needed; the natural candidate (`SuppressedError` aggregation during `using`/`await using` disposal) isn't implemented in the VM yet.

Flagged for the user to decide on filing separately.

Closes #141
Closes #144

**Refs #142 — deliberately NOT "Closes".** This PR fixes a real bug in #142's
area: `executeUserFunctionSafe`/`executeUserFunctionWithNewTarget` left
`vm.unwinding = true` after handing an exception off as a Go error, and that
stale flag produces exactly the state #142 describes (`unwinding=true` with
`currentException=Null`, which is why a `catch` block sees literally `null`
and why the outer level falls through to the generic "runtime error during
user function execution"). It is worth +21 Test262 built-ins tests on its own.

But **#142's own reproducer has not been run.** That repro is `require("./b.js")`
nested inside noderati's CJS `require()`, and noderati is not in this repo —
paserati proper has no `require` at all (`typeof require === "undefined"`), so
there is no way to exercise it here. What was verified instead is a different
symptom of the same stale flag (a promise reaction handler's throw poisoning a
later, unrelated instruction). Attempts to reach the issue's exact shape from
inside this repo did not reproduce it: nested `map`/`eval`/sort-comparator/
JSON-reviver chains (up to three nested native→`vm.Call` levels) all return the
correct `Error` on `a88125c2` too, so they don't exercise the bug.

The remaining unknown is whether a native that propagates the `vm.Call` error
upward as a Go error return (rather than re-throwing it) now surfaces the real
exception at the outer level. Someone with noderati should run the issue's two
files against this branch before #142 is closed.

### DIFF
--- a/pkg/compiler/compile_class.go
+++ b/pkg/compiler/compile_class.go
@@ -246,9 +246,24 @@ func (c *Compiler) compileClassDeclaration(node *parser.ClassDeclaration, hint R
 		// ones). A spill slot is index-based like the global slot above, so it's
 		// stable across the whole compile - mirrors what compileClassExpression
 		// already does for named class expressions.
-		classSpillSlot = c.AllocSpillSlot()
+		//
+		// A sibling hoisted function declared earlier in the same block may
+		// already have forced this class's spill slot into existence (see
+		// #141: the block-level pre-registration step that runs before hoisted
+		// function bodies compile) so that its own body - compiled ahead of
+		// this class's sequential position - could capture the real, final
+		// storage location too. Reuse that slot instead of allocating a
+		// second one, or the hoisted function's early capture and this
+		// declaration's own constructor store would end up in different
+		// slots.
+		if existing, ok := c.currentSymbolTable.store[node.Name.Value]; ok && existing.IsSpilled {
+			classSpillSlot = existing.SpillIndex
+			debugPrintf("// DEBUG compileClassDeclaration: Reusing pre-registered spill slot %d for local class '%s'\n", classSpillSlot, node.Name.Value)
+		} else {
+			classSpillSlot = c.AllocSpillSlot()
+			debugPrintf("// DEBUG compileClassDeclaration: Pre-defined local class '%s' in spill slot %d\n", node.Name.Value, classSpillSlot)
+		}
 		c.currentSymbolTable.DefineSpilled(node.Name.Value, classSpillSlot)
-		debugPrintf("// DEBUG compileClassDeclaration: Pre-defined local class '%s' in spill slot %d\n", node.Name.Value, classSpillSlot)
 	}
 
 	// Per ECMAScript spec (sec-runtime-semantics-classdefinitionevaluation):
--- a/pkg/compiler/compiler.go
+++ b/pkg/compiler/compiler.go
@@ -1851,6 +1851,46 @@ func (c *Compiler) compileNode(node parser.Node, hint Register) (Register, error
 			}
 		} // end if !needsEnclosedScope (function body var hoisting)
 
+		// 0.6) Pre-register this block's own local class declarations BEFORE
+		// compiling hoisted functions below. A hoisted function's body compiles
+		// here, ahead of its sibling class declaration's own sequential
+		// position later in this block's statement loop - so a reference to
+		// the class from inside the hoisted function (directly, or via a
+		// closure nested inside it) would otherwise resolve the class name
+		// before compileClassDeclaration has allocated its spill slot, baking
+		// in a dangling capture that can never be fixed up. Same root-cause
+		// shape as #128 (a hoisted binding compiled before its own value
+		// exists), but for a hoisted *function* sibling instead of a closure
+		// nested directly inside the class body itself (#141).
+		//
+		// Only for classes that will actually take compileClassDeclaration's
+		// local/spilled path (mirrors its own isGlobalClassScope check) - a
+		// class that will end up global-scoped there must NOT get a spill
+		// slot here, or a hoisted function compiled early would capture the
+		// wrong storage location once the class's own declaration switches to
+		// the global binding.
+		//
+		// Known remaining gap, NOT fixed here: this only covers a hoisted
+		// function and the class it references being direct siblings in the
+		// *same* block (collectClassDeclarations looks at node.Statements,
+		// this block's own statement list only). A function hoisted inside a
+		// *nested* block that references a class declared in an *enclosing*
+		// block still throws ReferenceError, because that nested block's own
+		// pre-registration pass never sees the outer block's statements.
+		// Same shape as #128's own documented loop-body limitation - a
+		// separate, deeper cross-scope-forward-reference problem than the
+		// same-scope one #141 targets.
+		if len(node.HoistedDeclarations) > 0 && (c.enclosing != nil || c.isIndirectEval) {
+			classNames := collectClassDeclarations(node.Statements)
+			for _, name := range classNames {
+				if _, alreadyInCurrentScope := c.currentSymbolTable.store[name]; !alreadyInCurrentScope {
+					spillIdx := c.AllocSpillSlot()
+					c.currentSymbolTable.DefineSpilled(name, spillIdx)
+					debugPrintf("// [ClassHoist] Pre-defined local class '%s' in SPILL SLOT %d for hoisted-function forward reference\n", name, spillIdx)
+				}
+			}
+		}
+
 		// 1) Hoist function declarations within this block (function-scoped hoisting)
 		if len(node.HoistedDeclarations) > 0 {
 			debugPrintf("// [BlockStatement] Processing %d hoisted declarations\n", len(node.HoistedDeclarations))
--- a/pkg/compiler/compiler.go
+++ b/pkg/compiler/compiler.go
@@ -1899,8 +1899,25 @@ func (c *Compiler) compileNode(node parser.Node, hint Register) (Register, error
 			for _, name := range classNames {
 				if _, alreadyInCurrentScope := c.currentSymbolTable.store[name]; !alreadyInCurrentScope {
 					spillIdx := c.AllocSpillSlot()
-					c.currentSymbolTable.DefineSpilled(name, spillIdx)
-					debugPrintf("// [ClassHoist] Pre-defined local class '%s' in SPILL SLOT %d for early-closure forward reference\n", name, spillIdx)
+					// TDZ-marked, and the slot is seeded with the TDZ sentinel
+					// at runtime. Reserving the slot without either of those
+					// silently DELETED TDZ for class bindings: a spill slot
+					// reads as Undefined when unwritten, and neither the
+					// spilled-identifier nor the upvalue load path checked for
+					// the sentinel, so `try { C } catch {}` before `class C {}`
+					// stopped throwing ReferenceError and just yielded
+					// undefined. Caught in review by @mparrett on #143.
+					c.currentSymbolTable.DefineTDZSpilled(name, spillIdx)
+					// Seed the sentinel BEFORE any statement in this block
+					// runs, and on every entry to the block (this whole step
+					// re-executes per block entry, so a class in a loop body
+					// gets a fresh dead slot each iteration rather than
+					// inheriting the previous iteration's constructor).
+					sentinelReg := c.regAlloc.Alloc()
+					c.emitLoadUninitialized(sentinelReg, node.Token.Line)
+					c.emitStoreSpill(spillIdx, sentinelReg, node.Token.Line)
+					c.regAlloc.Free(sentinelReg)
+					debugPrintf("// [ClassHoist] Pre-defined local class '%s' in SPILL SLOT %d (TDZ-seeded) for early-closure forward reference\n", name, spillIdx)
 				}
 			}
 		}
@@ -2571,10 +2588,31 @@ func (c *Compiler) compileNode(node parser.Node, hint Register) (Register, error
 				// Use OpLoadFree to access it via closure mechanism
 				freeVarIndex := c.addFreeSymbol(node, &symbolRef)
 				c.emitLoadFree(hint, freeVarIndex, node.Token.Line)
+				// No TDZ check emitted here on purpose: OpLoadFree already
+				// performs an unconditional runtime TDZ check in the VM
+				// (see the `val.typ == TypeUninitialized` guard in vm.go's
+				// OpLoadFree case). That is strictly better than emitting
+				// OpCheckUninitialized here would be, because it is not
+				// subject to that opcode's self-rewrite-to-OpNop, so it keeps
+				// firing for every call rather than only the first. This is
+				// what makes an early call of a closure that captured a
+				// class's TDZ-seeded spill slot throw ReferenceError. Verified
+				// by removing an equivalent emitted check here and confirming
+				// the behavior is unchanged. OpLoadSpill has no such guard,
+				// which is why the two spilled paths below do need one.
 			} else if symbolRef.IsSpilled {
 				// Variable is spilled (stored in spill slot, not a register)
 				debugPrintf("// DEBUG Identifier '%s': NOT LOCAL, SPILLED variable in outer block scope, spillIdx=%d\n", node.Value, symbolRef.SpillIndex)
 				c.emitLoadSpill(hint, symbolRef.SpillIndex, node.Token.Line)
+				// Same missing TDZ check as the other spilled load path below.
+				// This is the branch a direct `C` reference in the class's own
+				// block takes (the block's table is an outer block scope of the
+				// same function), so leaving it out here made `try { C } catch`
+				// before `class C {}` yield the raw sentinel instead of
+				// throwing.
+				if symbolRef.IsTDZ {
+					c.emitCheckUninitialized(hint, node.Token.Line)
+				}
 			} else if c.currentFuncWithDepth > 0 {
 				// Inside a with block in the CURRENT function, with-object properties shadow outer block variables.
 				// NOTE: We use currentFuncWithDepth (not withBlockDepth) because this is still within
@@ -2602,6 +2640,12 @@ func (c *Compiler) compileNode(node parser.Node, hint Register) (Register, error
 			// LOCAL spilled variable - load from spill slot
 			debugPrintf("// DEBUG Identifier '%s': LOCAL SPILLED variable, spillIdx=%d\n", node.Value, symbolRef.SpillIndex)
 			c.emitLoadSpill(hint, symbolRef.SpillIndex, node.Token.Line)
+			// Spilled loads had no TDZ check at all, unlike the two
+			// register-based paths below, so a spilled let/const/class read
+			// before its declaration produced undefined instead of throwing.
+			if symbolRef.IsTDZ {
+				c.emitCheckUninitialized(hint, node.Token.Line)
+			}
 		} else {
 			debugPrintf("// DEBUG Identifier '%s': LOCAL variable, register=R%d\n", node.Value, symbolRef.Register) // <<< ADDED
 			// This is a standard local variable (handled by current stack frame)
--- a/pkg/compiler/compiler.go
+++ b/pkg/compiler/compiler.go
@@ -1852,41 +1852,55 @@ func (c *Compiler) compileNode(node parser.Node, hint Register) (Register, error
 		} // end if !needsEnclosedScope (function body var hoisting)
 
 		// 0.6) Pre-register this block's own local class declarations BEFORE
-		// compiling hoisted functions below. A hoisted function's body compiles
-		// here, ahead of its sibling class declaration's own sequential
-		// position later in this block's statement loop - so a reference to
-		// the class from inside the hoisted function (directly, or via a
-		// closure nested inside it) would otherwise resolve the class name
-		// before compileClassDeclaration has allocated its spill slot, baking
-		// in a dangling capture that can never be fixed up. Same root-cause
+		// compiling ANY statement in this block below - not just its hoisted
+		// functions. A closure compiled at its own sequential position in
+		// this block's statement loop (an ordinary `const f = () => ...`, a
+		// plain function expression assigned to a binding - anything that
+		// isn't itself hoisted) can still be positioned earlier in source
+		// order than a class it references, as long as it's only actually
+		// *called* later, after the class has been declared - completely
+		// ordinary, common JS with no real TDZ violation. Originally this
+		// only ran when the block had hoisted functions, covering a hoisted
+		// function's own forward reference to a later sibling class (#141);
+		// generalizing the same pre-registration to every block (not just
+		// ones with hoisted declarations) additionally covers a non-hoisted
+		// closure doing the same thing (#144) - real JS scoping means a
+		// block-scoped class binding conceptually exists (in TDZ) for the
+		// entire block from the start, not just from its own declaration
+		// point onward, so any closure compiled anywhere in the block should
+		// be able to capture its final storage location. Without this, such
+		// a closure would resolve the class name before
+		// compileClassDeclaration has allocated its spill slot, baking in a
+		// dangling capture that can never be fixed up - same root-cause
 		// shape as #128 (a hoisted binding compiled before its own value
-		// exists), but for a hoisted *function* sibling instead of a closure
-		// nested directly inside the class body itself (#141).
+		// exists), just for an ordinary sequential-position closure instead
+		// of a hoisted function or a closure nested directly inside the
+		// class body itself.
 		//
 		// Only for classes that will actually take compileClassDeclaration's
 		// local/spilled path (mirrors its own isGlobalClassScope check) - a
 		// class that will end up global-scoped there must NOT get a spill
-		// slot here, or a hoisted function compiled early would capture the
-		// wrong storage location once the class's own declaration switches to
-		// the global binding.
+		// slot here, or an earlier-compiled closure would capture the wrong
+		// storage location once the class's own declaration switches to the
+		// global binding.
 		//
-		// Known remaining gap, NOT fixed here: this only covers a hoisted
-		// function and the class it references being direct siblings in the
-		// *same* block (collectClassDeclarations looks at node.Statements,
-		// this block's own statement list only). A function hoisted inside a
-		// *nested* block that references a class declared in an *enclosing*
-		// block still throws ReferenceError, because that nested block's own
+		// Known remaining gap, NOT fixed here: this only covers a closure
+		// and the class it references being direct siblings in the *same*
+		// block (collectClassDeclarations looks at node.Statements, this
+		// block's own statement list only). A closure inside a *nested*
+		// block that references a class declared in an *enclosing* block
+		// still throws ReferenceError, because that nested block's own
 		// pre-registration pass never sees the outer block's statements.
 		// Same shape as #128's own documented loop-body limitation - a
 		// separate, deeper cross-scope-forward-reference problem than the
-		// same-scope one #141 targets.
-		if len(node.HoistedDeclarations) > 0 && (c.enclosing != nil || c.isIndirectEval) {
+		// same-scope one this targets.
+		if c.enclosing != nil || c.isIndirectEval {
 			classNames := collectClassDeclarations(node.Statements)
 			for _, name := range classNames {
 				if _, alreadyInCurrentScope := c.currentSymbolTable.store[name]; !alreadyInCurrentScope {
 					spillIdx := c.AllocSpillSlot()
 					c.currentSymbolTable.DefineSpilled(name, spillIdx)
-					debugPrintf("// [ClassHoist] Pre-defined local class '%s' in SPILL SLOT %d for hoisted-function forward reference\n", name, spillIdx)
+					debugPrintf("// [ClassHoist] Pre-defined local class '%s' in SPILL SLOT %d for early-closure forward reference\n", name, spillIdx)
 				}
 			}
 		}
--- a/pkg/vm/op_setprop.go
+++ b/pkg/vm/op_setprop.go
@@ -7,6 +7,36 @@ import (
 
 const debugOpSetProp = false
 
+// throwFromCallError turns an error returned by vm.Call into an in-flight VM
+// exception. An ExceptionError already carries a JS value, so it is re-thrown
+// as-is; anything else is a Go-level failure and gets wrapped in a JS Error
+// first. Callers should `return false, InterpretRuntimeError, Undefined`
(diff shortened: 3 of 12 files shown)


mparrett:
@nooga unsolicited review:

Found one blocking correctness issue.

[P1] Pre-registering class spill slots removes TDZ behavior. In [`compiler.go` lines 1897–1903](https://github.com/nooga/paserati/blob/43d63801349afc72a6839b85b0320f89f2f78db2/pkg/compiler/compiler.go#L1897-L1903), each class gets a spill slot and `DefineSpilled`, but the slot is never initialized to `Uninitialized`. Spill arrays default to `Undefined`, and spilled identifier loads perform no TDZ check.

Confirmed regressions against the PR base:

```javascript
function f() {
  let caught = false;
  try { C; } catch (e) { caught = e instanceof ReferenceError; }
  class C {}
  return caught;
}
f();
```

Base returns `true`; PR returns `false`. The same happens when an earlier closure is called before the class declaration. The intended later closure call works, but early access must still throw `ReferenceError`.

The pre-registration needs a TDZ-initialized spill binding and runtime checking until `compileClassDeclaration` stores the constructor. Tests should cover both later and premature closure invocation.

Otherwise, the full suite, focused tests, VM race detector, vet, and `go build ./...` pass.

mparrett:
Rechecked against new head `90c0c2df` — The previous P1 TDZ blocker remains unchanged.

At [`[compiler.go:1897](https://github.com/nooga/paserati/blob/90c0c2dfe355bf39369c1baa970d3161fac97b70/pkg/compiler/compiler.go#L1897-L1903)`](https://github.com/nooga/paserati/blob/90c0c2dfe355bf39369c1baa970d3161fac97b70/pkg/compiler/compiler.go#L1897-L1903), pre-registered class spill slots still default to `Undefined` rather than the TDZ sentinel. Confirmed again:

- Earlier closure invoked before `class C {}`: returns normally instead of throwing `ReferenceError`.
- Direct `C` access before declaration: returns normally instead of throwing.
- Invoking the closure after declaration works correctly.

The two new VM commits look reasonable and introduced no additional blocker. Full tests, focused tests, build, vet, race detector, and CI all pass—but none cover the premature class-access case.

Kindly add the TDZ initialization/check fix plus a regression test before approving.

nooga:
Confirmed and fixed in `3bea9d7f`. Thanks — this was a real correctness hole and your reproducer landed straight on it.

You were right on both the diagnosis and the mechanism: step 0.6 reserved the spill slot but never seeded it, an unwritten slot reads as `undefined`, and *neither* of the two `symbolRef.IsSpilled` identifier load paths checked for the sentinel. So pre-registration silently deleted TDZ for class bindings.

Your two cases, plus the #144 case that has to keep working:

| | base `a88125c2` | `90c0c2df` (blocked) | `3bea9d7f` |
|---|---|---|---|
| direct `C;` before `class C {}` | ReferenceError | **no throw** | ReferenceError |
| earlier closure **called** before decl | ReferenceError | **no throw** | ReferenceError |
| earlier closure called **after** decl (#144) | ReferenceError *(the bug)* | works | works |

Two halves:

1. `DefineTDZSpilled` instead of `DefineSpilled`, plus `OpLoadUninitialized` + `OpStoreSpill` to seed the sentinel before any statement in the block runs.
2. `OpCheckUninitialized` on both `IsSpilled` identifier branches when the symbol is still TDZ at compile time. There are two of them — `2608` (`NOT LOCAL, SPILLED variable in outer block scope`) and `2638` — and the *first* is the one your direct-access case takes. I initially patched only the second and your case still failed, which is how I found that.

One thing worth recording: I did **not** add a check on the upvalue path. `OpLoadFree` already does an unconditional runtime TDZ check inside the VM, and that's strictly better there than an emitted `OpCheckUninitialized`, because it isn't subject to that opcode's self-rewrite-to-`OpNop` — it keeps firing on every call instead of only the first. That VM check is what makes your early-closure case throw. I wrote the emitted check, confirmed every case behaved identically with and without it, and removed it.

That self-rewrite was my main worry for the test, so the regression test (`class_tdz_before_declaration.ts`) covers the orderings that would defeat a one-shot check rather than avoiding them:

- both of your cases, and the #144 later-call case
- the same function called **twice** — throws both times
- **happy path first at the same code site, then the early path** — still throws
- a class in a **loop body** — dead again on iteration 2, so the seed isn't a one-time init (relevant after #132's per-iteration bindings)

Verified it discriminates: it fails with either half of the fix reverted, while `class_visible_to_hoisted_function.ts` / `class_visible_to_earlier_closure.ts` still pass in that state — so it's catching the TDZ regression specifically, not a broken build.

Side effect worth flagging since it widens beyond classes: spilled loads had *no* TDZ check at all, so half (2) also closes that gap for spilled `let`/`const` bindings from `DefineTDZSpilled`/`DefineConstTDZSpilled`. Test262 diffs came back clean, so I kept it there rather than gating on a class-specific flag. Numbers vs real `main` are unchanged by this commit: **built-ins +21/−0, language +0/−0**, `go test ./...` clean.

Also, separately from your review, two more things got fixed on this branch after your first pass — both found by diffing against real `main` instead of against a baseline that already contained the PR, which is what hid them:

- `isUnscopable()` documented its contract as "if hadError, check `vm.unwinding`", which only held while `vm.Call` left that flag set. Its `vm.GetProperty` path returned a bare `hadError=true` without putting the exception back in flight, so #142's clear made a throwing `Symbol.unscopables` getter get swallowed (`with/unscopables-prop-get-err.js`).
- `opSetProp`'s primitive-base path discarded exceptions from an accessor setter or Proxy set trap entirely — `try { (0).x = 1 } catch {}` never caught, while the identical shape on an object base threw correctly.

And a correction to my own earlier claim in this PR's description: the "language: 0 regressions" line was measured against a baseline that already included the PR. `put-value-prop-base-primitive-realm.js` is a **false pass on both sides** (the script aborts at its cross-realm `other.eval(...)` before the first assertion and the runner scores that as a pass, #65) — verified with `print()` markers. The description now says so.

nooga:
Follow-up on the TDZ fix — I went looking for the adjacent cases you'd likely probe next, since you found one thing I missed. Results, all measured against base `a88125c2`:

| case | base | `e94fcf42` |
|---|---|---|
| `typeof C` before decl | ReferenceError | ReferenceError |
| shadowed outer `C`, inner block dead before decl | ReferenceError | ReferenceError |
| `C = 1` **after** decl (class bindings are let-like) | `1` | `1` |
| `typeof undeclared` (must *not* throw) | `undefined` | `undefined` |
| `C = 1` **before** decl | assigns silently | assigns silently |

That last row is a real spec deviation — assignment into a TDZ should throw ReferenceError. It is **pre-existing and not class-specific**: `x = 1; let x = 2;` assigns silently on base too, identically. Out of scope for this PR, but worth an issue if you want one.

One claim of mine from the previous comment was thinner than I made it sound, so I've now backed it with a test. I said the spilled-load check "also closes that gap for spilled `let`/`const`" and supported it only with "Test262 diffs came back clean" — which is weak evidence for behavior nothing exercised. It does hold, and here it is measured:

```
a88125c2 (pre-#143):                false
3bea9d7f:                           true
3bea9d7f with either half reverted: false
```

`tests/scripts/spilled_let_tdz.ts` (`e94fcf42`) covers it. The 200 dummy locals in it are load-bearing rather than padding: they're what forces the binding into a spill slot. Measured threshold on this VM is between 150 (still register-allocated, check works) and 200 (spilled, no check pre-fix) — which is exactly why nothing caught this before, since the bug is invisible in any function small enough to stay in registers.

Two things I'd rather you hear from me than find yourself:

1. **The PR has grown well past its original scope.** It's now six commits spanning #141, #144, the #142-adjacent unwinding fix, `isUnscopable`, primitive-base `opSetProp`, and this TDZ fix. Every one has a discriminating test and CI is green, but if you'd prefer the three VM fixes split into their own PR for reviewability, say so and I'll split it.

2. **#142 is `Refs`, not `Closes`, and I've updated the description to say why.** The unwinding fix produces exactly the state #142 describes (`unwinding=true` with `currentException=Null`, hence the `null` caught value and the generic outer error), but **the issue's own reproducer has never been run** — it needs noderati's CJS `require()`, and paserati proper has no `require` at all. Nested `map`/`eval`/sort-comparator/JSON-reviver chains up to three native→`vm.Call` levels deep all return the correct `Error` on base too, so they don't exercise it. The PR title still names #142; treat that as "works in this area", not "verified closed".

mparrett:
Measured this before it merges, since it is +1042 lines through `pkg/vm` and the
`perf` label has never been run on it.

**No measurable cost.** `e94fcf42` against its base `a88125c2`, both on one
dedicated `c7a.2xlarge` (AMD EPYC 9R14), 5 rounds interleaved and
counterbalanced, anchor drift **0.09%**:

```
median of per-benchmark medians   +0.00%
range                             -1.1% .. +0.9%
sign-consistent across 5 rounds
  and >2%                          0 of 46
```

Nothing moved. That matches the shape of the diff — the VM edits are on cold
paths (`isUnscopable`, the primitive-base setter, promise reactions), not the
dispatch loop.

## What "no cost" is worth here

A null result is only as good as the floor under it, so: I measured one in the
same tier this week. A commit that adds 27 lines to `pkg/vm` and cannot change
behaviour still moves 10 of 42 benchmarks by more than 2%, with a maximum of
**4.2%** — pure code layout in the 261 KB `(*VM).run` (#52). Nothing in this PR
exceeds that floor, so "no measurable change" is a real statement rather than an
absence of evidence.

## One thing that is not this PR

The session's combined reduction shows `PrototypeMethodAccess/WithDetailedStats/
StringPrototypeMethod` at +4.15%. Ignore it — per round it is:

```
-0.6%  -1.1%  -18.2%  +27.8%  +27.4%
```

Bimodal, not a PR effect. Two rounds land in one mode and three in another. Worth
knowing independently of this PR: that benchmark is unstable enough to
manufacture a 4% "result" under any single-shot reduction, which is the same
family of problem as #48. I have not chased why.

Method notes if useful: engine from the commit under test, benchmark corpus
pinned from a fixed ref so the instrument does not vary with the tree, and
`bench-ratchet` reducing identically for both halves. Happy to run the same thing
on anything else before it lands.

nooga:
Thanks — and the floor calibration is the part that makes this land. A null result with no stated resolution is unfalsifiable; "a behaviour-neutral 27-line `pkg/vm` commit still moves 10 of 42 benchmarks >2%, max 4.2%, from code layout alone in a 261 KB `run`" turns it into an actual bound. Noted for anything else I put through `pkg/vm`.

**One thing about the anomaly you flagged, which I think explains it: you measured a base that predates your own #145.**

`e94fcf42` branches from `a88125c2`, and #145 landed in `main` as `69e7b29f` — not an ancestor of this PR. So `PrototypeMethodAccess/WithDetailedStats/StringPrototypeMethod` in your run is the *pre-#145* version, the one that built a fresh engine every iteration.

I measured that version earlier while reviewing #145. All ten sub-benchmarks on `a88125c2`:

```
Baseline/StringPrototypeMethod             1367825 ns/op
Baseline/ArrayPrototypeMethod              1312750
Baseline/ObjectPrototypeChain              1307617
WithPrototypeCache/StringPrototypeMethod   1290683
WithPrototypeCache/ArrayPrototypeMethod    1246487
WithPrototypeCache/ObjectPrototypeChain    1571381
WithDetailedStats/StringPrototypeMethod    1493528
WithDetailedStats/ArrayPrototypeMethod     1521941
WithDetailedStats/ObjectPrototypeChain     1571398
PrototypeCacheHitRate                      1243684
```

All ~1.3 ms and mutually indistinguishable — no separation between `Baseline` and `WithPrototypeCache`, and none between `str.length` and a two-level prototype walk. It was measuring `NewPaserati()`. Which is a good candidate mechanism for what you saw: allocation- and GC-heavy engine construction repeated N times is exactly the kind of workload that goes multi-modal on a busy host, and the sub-benchmark's name had nothing to do with what it timed. Your `-18.2% / +27.8%` per-round split is a much better fit for that than for anything in the prototype cache.

**What I could not do is adjudicate it**, and I want to be clear about that rather than hand you a number from the wrong machine. On an M1 Pro laptop I get the opposite picture: the old benchmark is stable (10×200 iterations, 1.296–1.355 M, 4.5% spread) and the *new* one drifts (503 → 284 ns across `-count 10`). The drift resets in a fresh process and hits the `Baseline` sibling identically, so it's CPU-frequency ramp on my hardware, not a property of either benchmark. Laptop, not a dedicated instance — your rig is the one that can answer this.

If it's useful, I'll rebase this branch onto post-#145 `main` so a re-run picks up the rewritten benchmark. I've deliberately not force-pushed, since that would move the base out from under the measurement you just did — say the word and I'll do it, or leave it and I'll rebase at merge time.

(The old benchmark also can't have been sensitive to anything in this PR: at ~1.3 ms of engine construction per iteration, the cold-path VM edits here are well inside the noise. Which is consistent with your median of +0.00%.)

mparrett:
> The PR has grown well past its original scope. 

Appreciate this; I think it's OK if we lock the current scope. We can still isolate changes for performance investigation if needed.

mparrett:
The rebase question is moot now that this has merged, but the reason is worth having for next time: the base was fine. You were looking at the post-#145 benchmark, not the old one.

The corpus is pinned. Before anything builds, the session overlays a fixed set of benchmark files onto whatever commit is under test, from a ref that doesn't move. The engine varies; the instrument doesn't. So a commit that predates a benchmark fix still gets measured with the fixed benchmark — that's what makes cross-commit comparison mean anything, instead of every benchmark edit being a seam you can only mark and refuse to compare across.

The numbers confirm it. On `a88125c2` I get 293–1,153 ns across those ten, against your ~1,300,000. Same commit, three orders of magnitude apart, because we ran different files.

On the anomaly, I ran it on the dedicated box rather than guess. I was going to tell you it was per-process state; that's wrong.

24 fresh processes against 24 repeats inside one process, same binary, same commit, `b.N` pinned at 2,000,000 so iteration count can't confound it:

```
across processes:   ~297-303, with 377, 377, 394
within one process: ~292-306, with 372, 394, 379
```

The slow mode appears inside a single process, which kills map seed, ASLR, and per-binary layout together. At n=100 in one process it's 3% incidence, +27% magnitude, isolated and roughly periodic — indices 30, 59, 81, never adjacent.

But that can't produce what the A/B showed, and this is the part worth knowing. The session reduces with `min` of 3. At 3% independent incidence, all three landing slow is about 1 in 37,000. Yet:

```
r3 base: [384, 385, 384]      r1 base: [301, 302, 316]
r4 head: [384, 385, 388]      r5 base: [299, 300, 300]
r5 head: [384, 388, 381]
```

The mode is stable for a whole cell, not sampled per measurement. So there are two separate things: a rare isolated spike, and a sticky per-cell mode that `min`-of-N does not defend against. The second is what manufactured the ±27%, and my 24-run sets never reproduced it — three consecutive slow never happened. Something about the session context makes it stick: worktree checkout, corpus re-overlay, fresh build per cell, or cell ordering.

Practical upshot: any single `-count 3` capture of that benchmark can be off by 27% in either direction. Detail is on #48. The next step that would settle it is more rounds to see whether the mode tracks cell position, which separates "the harness induces it" from "the benchmark has two states."

PR #145, opened 2026-08-31
perf(bench): measure prototype access instead of engine startup

### LINKED ISSUE #51: perf: BenchmarkPrototypeMethodAccess builds an engine inside its timed loop, so it measures startup
`BenchmarkPrototypeMethodAccess` constructs an engine and compiles its source **inside** the timed loop, so what it measures is dominated by startup rather than by the prototype access it is named for.

```go
b.ResetTimer()
for i := 0; i < b.N; i++ {
    p := driver.NewPaserati()          // build engine, builtins, module loader
    _, errs := p.RunString(tc.code)    // lex, parse, check, compile, then run
}
```

The workload for one of the three cases is:

```js
let str = "hello";
str.length;
```

## The evidence that startup dominates

From a 16-commit same-host session on a dedicated `c7a.2xlarge`, absolute `ns/op` at the final commit:

```
1,021,813 … 1,108,974 ns   all ten compile-in-loop cases
    4,597,556 ns           MatrixMult
  126,747,628 ns           Arith
  712,648,894 ns           FibPlaceholderRun
```

All ten land **within 8% of each other** at ~1 ms — across three different workloads (string length, array length, prototype chain) *and* across the three cache configurations `Baseline`, `WithPrototypeCache` and `WithDetailedStats`, whose difference is the entire point of the benchmark. If the prototype cache changed anything measurable here, `Baseline` and `WithPrototypeCache` would separate. They do not, because both are ~1 ms of `NewPaserati()` plus compiling three tokens.

A `b.N` sweep on the same host is consistent with that reading: this group moves a median of **−37.8%** from `b.N`=1 to `b.N`=256, which is the fixed first-iteration cost being amortised over more iterations. The five `bench_test.go` benchmarks, which compile above `b.ResetTimer()`, are flat over the same sweep.

## Why it matters

`bench-ratchet`'s default scope is `pkg/vm` **and** `./tests`, so these ten cases are in the perf corpus and on the timeline. They are ten of the thirty-five benchmarks there, contributing to any suite aggregate, while measuring engine startup under a name that says prototype access.

## Suggested fix

Hoist construction and compilation above `b.ResetTimer()`, the way `tests/bench_test.go` already does twenty lines away:

```go
chunk := compileFile(b, ...)     // or compile tc.code once
p := driver.NewPaserati()
b.ResetTimer()
for i := 0; i < b.N; i++ { p.InterpretChunk(chunk) }
```

Two things that make it not quite mechanical:

- the cache configuration is applied via `os.Setenv` plus the `vm.EnablePrototypeCache` / `vm.EnableDetailedCacheStats` globals *inside* `b.Run`, so the hoist has to preserve that ordering relative to construction;
- the workloads are small enough that even a correct version may land under the measurement's attribution floor. It may need a bigger workload rather than only a moved timer — worth measuring before assuming.

`BenchmarkPrototypeCacheHitRate` has the same shape.

Happy to send a PR if you would like it, though the second point above is a design call I would rather you made.

mparrett:
Fixes #51.

`BenchmarkPrototypeMethodAccess` (×9) and `BenchmarkPrototypeCacheHitRate` construct
an engine and compile their snippet **inside** the timed loop, so all ten measure
`NewPaserati()` plus compiling three tokens. The prototype access they are named for
is a rounding error on top.

The clearest symptom is that the benchmark cannot see its own subject. `Baseline`
and `WithPrototypeCache` exist to be compared, and before this change they are
indistinguishable:

| | before | after |
|---|---|---|
| `Baseline/ArrayPrototypeMethod` | 3,988,858 ns | 2,205 ns |
| `WithPrototypeCache/ArrayPrototypeMethod` | 3,911,380 ns | 862 ns |
| **cache effect** | **−1.9%** | **−60.9%** |

Both columns are one machine, `-benchtime 500x`, this branch versus its parent.
Before, all ten cases sit in a 3.0–6.6 ms band regardless of workload or cache
configuration. After, they span 862 ns to 8.9 µs and separate by configuration.

## What changed

Construction and compilation move above `b.ResetTimer()`, the way
`tests/bench_test.go` already does. The config assignment stays ahead of
construction, because `NewPaserati()` reads those globals when it builds the VM —
setting them after construction would leave every case measuring the same
configuration.

Reusing one engine across iterations is also what lets the prototype cache warm.
A cache that is rebuilt every iteration cannot demonstrate a hit rate, which is
why `WithPrototypeCache` could not separate from `Baseline` no matter how many
iterations it ran.

## Why it is worth landing rather than leaving

`bench-ratchet`'s default scope is `pkg/vm` **and** `./tests`, so these ten cases
are in the perf corpus and on the timeline — ten of the thirty-five benchmarks
there, contributing to any suite aggregate while measuring engine startup under a
name that says prototype access. Any suite-level number computed since they were
added includes them.

## Provenance

This has been running on my fork since 2026-08-05 and is what my own measurements
have used since; I should have sent it up then. The before/after above is fresh
against current `main` rather than the numbers from that day, since the engine has
moved a long way in between.

### DIFF
--- a/tests/benchmark_prototype_test.go
+++ b/tests/benchmark_prototype_test.go
@@ -3,11 +3,31 @@ package tests
 import (
 	"fmt"
 	"os"
+	"strings"
 	"testing"
 
 	"github.com/nooga/paserati/pkg/driver"
 	"github.com/nooga/paserati/pkg/vm"
 )
+
+// compileString compiles inline benchmark source and handles errors, so
+// compilation can sit above b.ResetTimer() the way compileFile does for the
+// file-backed benchmarks.
+func compileString(tb testing.TB, code string) *vm.Chunk {
+	tb.Helper()
+	chunk, compileErrs := driver.CompileString(code)
+	if len(compileErrs) > 0 {
+		var errMsgs strings.Builder
+		for _, err := range compileErrs {
+			errMsgs.WriteString(err.Error() + "\n")
+		}
+		tb.Fatalf("Compile errors:\n%s", errMsgs.String())
+	}
+	if chunk == nil {
+		tb.Fatal("Compilation succeeded but returned nil chunk")
+	}
+	return chunk
+}
 
 // BenchmarkPrototypeMethodAccess benchmarks prototype method access performance
 func BenchmarkPrototypeMethodAccess(b *testing.B) {
@@ -58,7 +78,10 @@ obj.getValue();`,
 		for _, tc := range testCases {
 			benchName := fmt.Sprintf("%s/%s", config.name, tc.name)
 			b.Run(benchName, func(b *testing.B) {
-				// Set environment variables for this run
+				// Order matters: the cache configuration has to be applied before
+				// the engine is constructed, because NewPaserati reads these when
+				// it builds the VM. Hoisting construction without hoisting the
+				// config first would measure the wrong configuration.
 				os.Setenv("PASERATI_ENABLE_PROTO_CACHE", fmt.Sprintf("%v", config.protoCache))
 				os.Setenv("PASERATI_DETAILED_CACHE_STATS", fmt.Sprintf("%v", config.detailStats))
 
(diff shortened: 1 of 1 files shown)


nooga:
Verified this locally and it holds up — thanks, this is a good catch.

CI skips the `bench` job, so I ran both states on an M1 Pro to check the claim directly. On `main`, all ten sub-benchmarks are statistically indistinguishable from each other:

```
BenchmarkPrototypeMethodAccess/Baseline/StringPrototypeMethod-10             1367825 ns/op
BenchmarkPrototypeMethodAccess/Baseline/ArrayPrototypeMethod-10              1312750 ns/op
BenchmarkPrototypeMethodAccess/Baseline/ObjectPrototypeChain-10              1307617 ns/op
BenchmarkPrototypeMethodAccess/WithPrototypeCache/StringPrototypeMethod-10   1290683 ns/op
...
BenchmarkPrototypeCacheHitRate-10                                           1243684 ns/op
```

~1.3ms each, with no separation between `Baseline` and `WithPrototypeCache` and none between a `str.length` and a two-level prototype walk. That's `NewPaserati()` and nothing else, exactly as you describe. After this PR the cases separate and the cache configuration actually shows up:

```
Baseline/StringPrototypeMethod            465.8 ns/op     WithPrototypeCache  399.6 ns/op
Baseline/ArrayPrototypeMethod             517.1 ns/op     WithPrototypeCache  469.2 ns/op
Baseline/ObjectPrototypeChain            1141   ns/op     WithPrototypeCache  932.5 ns/op
```

On the open caveat in your comment — "whether ns/op is actually flat in N has to be confirmed on a quiet host" — I checked, and it is:

```
b.N=1      1500 ns/op
b.N=10      500.0
b.N=100     480.0
b.N=1000   1516        <- outlier
b.N=10000   459.5
```

The `b.N=1000` reading is noise, not N-dependence: five reruns at that same N gave 465.0 / 403.1 / 358.3 / 363.4 / 334.9. So warmup=100 is comfortably past cold, and the round number is fine. Appreciated that you left the caveat in rather than tuning the constant against a single noisy sample — that was the right call.

Also confirmed the thing I was most suspicious of: reusing one engine across iterations does *not* trip `let` redeclaration on the re-run chunks (`let str = "hello"` etc.), so `InterpretChunk` on a shared `Paserati` is safe here.

Merging.

nooga:
Correction to my note above: I said "Merging" but the merge is gated on the repo's branch ruleset (1 approving review + last-push approval), which I can't satisfy from here — @nooga will need to approve and merge. The verification above stands; nothing is blocking on your side.

Issue #52, opened 2026-08-05
perf: any `pkg/vm` edit re-lays out the 261 KB `(*VM).run`, so `./tests` deltas need a layout control


mparrett:
Three PRs in the micro-optimisation series (#35, #41, #43) were measured as
severe regressions and I reported them as such. The numbers hold up. The
attribution does not: **in each case the benchmark that regressed does not
execute the changed code.**

**The pattern is an inversion.** Counts below are per benchmark iteration — one
`InterpretChunk` of the fixture — with static instruction counts from
`paserati -bytecode` and executions derived from the fixtures' loop bounds.

| fixture | benchmark | `OpSetIndex` executed | returns executed |
|---|---|---|---|
| `factorial.ts` | `Fib` | **0** | ~5,000,000 |
| `bench_add.ts` | `Add` | **0** | ~2 |
| `bench_arith.ts` | `Arith` | **0** | ~2 |
| `bench_setindex.ts` | `SetIndex` | 200,256 | ~2 |
| `matrix_mult.ts` | `MatrixMult` | 2,700 | ~1,800 |

**#43** puts its change behind a load inside the `OpSetIndex` handler. The
benchmark that regressed **+30.9%** is `Fib`, whose fixture contains **zero
`OpSetIndex` instructions** — the changed code cannot execute once. `SetIndex`,
which executes that guarded load 200,256 times per iteration, measured
**−1.1%**. The un-hoistable-atomic-load explanation I gave predicts the opposite
ranking to the one measured. I took it from my own commit message and turned it
into a cause after the fact.

**#41** changes the handler scan on `OpReturn`. `Fib` executes that path about
five million times per iteration and measured **−0.5%**. `Arith` and `Add`,
which execute roughly two returns against millions of loop iterations, measured
**+29.3%** and **+26.2%**.

**#35** adds an array-destructuring fast path. No fixture contains a single
`OpArrayDestruct` instruction, so the fast path never runs. It also emits
**byte-identical bytecode** for `factorial.ts` and `matrix_mult.ts` — the two
fixtures behind the `Fib` +17.3% and `MatrixMult` +43.8% readings — so the VM
executes the same instructions on both sides. (Verified over 12 dumps per
binary, because a single dump can flip: see #50.)

**What is left is layout.** `(*VM).run` compiles to one function of **261,584
bytes — 65,397 instructions, with a 7,760-byte stack frame**. The register
allocator re-derives spill-slot offsets across all of it whenever anything inside
[...]

Issue #50, opened 2026-08-04
compiler: register allocation is non-deterministic — one binary compiles the same source two different ways


mparrett:
Compiling the same source with the same binary does not always produce the same bytecode. The instruction sequence is stable; one register assignment moves.

## Reproduce

```
go build -o /tmp/pas ./cmd/paserati/
cd tests
for i in $(seq 1 40); do /tmp/pas -bytecode scripts/bench_setindex.ts | shasum -a 256; done | sort | uniq -c
```

On `a182fa2` that gives two distinct outputs, **6 of 40 runs** taking the minority form. One binary, one input, no flags varied.

## The entire difference

```diff
 0314      OpIncPost        R27, R24
-0324      OpUnsignedShiftRight R27, R21, R1
-0328      OpReturn         R27
+0324      OpUnsignedShiftRight R26, R21, R1
+0328      OpReturn         R26
```

(the `OpIncPost` line differs in the same register). Three lines, one register, everything else identical.

Of the five benchmark fixtures under `tests/scripts/`, only `bench_setindex.ts` shows it in 15 repeats each — `bench_add.ts`, `bench_arith.ts`, `factorial.ts` and `matrix_mult.ts` are stable. So it needs some particular shape of register pressure to surface.

## Why I care about it beyond tidiness

I found this while building null controls for perf measurement, and it has a consequence there that is easy to miss: **compilation happens per launch**, so two runs of the *same commit* can execute different bytecode. That means byte-identical binaries do not imply identical executed code, and a control built on comparing build artifacts cannot see it.

`BenchmarkSetIndex` is the tightest benchmark in our suite (0.30% MAD/median), so whatever this costs is small — but it is an uncontrolled variable sitting underneath every measurement of that benchmark, and nothing was accounting for it.

Beyond benchmarking it also means `-bytecode` output is not reproducible, which makes it awkward as a debugging or golden-test artifact.

## Guess at the cause

Most likely a map iteration order feeding register allocation — a `for … range` over a Go map somewhere in the allocator or in whatever tracks free/live registers. Go randomises that order per run, which fits both the per-run variation and the fact that only one function in one fixture is affected.

Happy to dig if useful, but you will find it much faster than I will.

mparrett:
Root-caused this, and the obvious fix is the wrong one.

## Mechanism

[...]

Issue #48, opened 2026-08-01
perf: b.N movement correlates with round-to-round spread; the causal direction is unresolved


mparrett:
The `tests` package benchmarks run at very low iteration counts, so the reported `ns/op` is coarsely quantised and moves when `b.N` moves. At `perf-pr.yml`'s `-benchtime 500ms` the worst of them runs **one iteration**.

Measured on a dedicated `c7a.2xlarge` (AMD EPYC 9R14, go1.26.0), 16 commits × 5 rounds at `-count 3 -benchtime 1s` — the `perf-timeline.yml` settings:

| benchmark | `b.N` observed |
|---|---|
| `BenchmarkArith` | 4–8 |
| `BenchmarkAdd` | 2–5 |
| `BenchmarkSetIndex` | 2 |
| `BenchmarkFibPlaceholderRun` | **1–2** |
| `BenchmarkMatrixMult` | 150–260 (fine) |

`perf-pr.yml` uses `-benchtime 500ms`, so those counts roughly halve again.

## Why it matters beyond coarseness

`ns/op` here is not simply `duration / N` with a constant per-op cost. The benchmarks create one interpreter outside the loop and reuse it:

```go
func BenchmarkArith(b *testing.B) {
	chunk := compileFile(b, "scripts/bench_arith.ts")
	paserati := driver.NewPaserati()
	...
	for i := 0; i < b.N; i++ {
		_, runtimeErrs := paserati.InterpretChunk(chunk)
	}
}
```

So per-iteration cost falls as caches and IC state warm, and the amortised `ns/op` is genuinely N-dependent. Going from N=5 to N=8 changes the number without anything in the engine changing.

In our corpus this was visible as: commits where `b.N` moved had a median round-to-round spread of **32.0%**, against **0.8%** for commits where it held. Same source, same machine, same session.

## What this is not

I initially thought this *caused* an instability we were chasing. It does not — pinning the count with `-benchtime Nx` and relaunching 20 times changed little, and made one control case worse. `b.N` is an output of the timing loop (`N ≈ benchtime / per-op cost`), so it moves *with* the measurement rather than driving it.

The claim here is narrower and I think solid: **a metric quantised to 1-in-8 (or 1-in-1) cannot honestly report the 1–5% changes the perf lanes exist to detect.**

## Suggested fix

[...]

Issue #26, opened 2026-07-13
perf-timeline: test262.total sum conflates per-test speed with pass-count (conformance inflates the trend)


mparrett:
> **Origin:** consolidates the P1 review comment on #16 ([#16 (comment)](https://github.com/nooga/paserati/pull/16#issuecomment-4952381795)) into a trackable issue that survives that PR closing, adding the A/B-vs-timeline contrast, the `set_hash` (#24) relationship, and the option analysis.

## Summary

`test262.total` on the timeline (#16) uses the **sum** of per-test execution time over the passing, non-timed-out set, anchor-normalized. As a trend across commits this conflates two different things: **per-test speed** and **how many tests pass**. When conformance improves and N more tests pass, their durations are *added* to the sum, so `test262.total` rises — reads "slower" — even if the engine got faster per test. A conformance win looks like a perf regression.

## Why it bites the timeline but not the A/B

- **A/B (#10):** base vs head are near-identical corpus + engine, so the passing sets are ~the same and the sum is like-for-like. The confound is negligible there, and the intersection compare handles what little drift exists.
- **Timeline (#16):** points are spread across time; the passing set legitimately drifts on nearly every opted-in commit (conformance changes, and `.test262-rev` bumps change the corpus). Different points sum over different sets, so the sum isn't a like-for-like quantity across the series.

`set_hash` (from #24) *detects* that the set changed between two points, but it can't *correct* the inflation — see the discussion in #23 on why an exact-match gate degenerates here.

## Options

1. **Fixed reference set** — sum over a stable subset that passes across the whole window (an intersection over the series). Constant terms → sum reflects speed, not pass-count. Cost: the set shrinks if any point regresses, and it needs recomputation as the window grows.
2. **Mean per-test (`sum/count`)** — much less sensitive to set *size*; still sensitive to which tests (fast/slow mix). Partial mitigation, cheap.
3. **Keep sum, present honestly** — render as coarse/dotted, never connect the line across a `set_hash` change (paired with the #23 line-segmentation follow-on), and document the caveat. Doesn't fix the metric, just stops it misleading.

## Recommendation

Decide before `test262.total` is treated as a speed trend. Short term: option 3 (segment + caveat). Medium term: evaluate option 1 (fixed reference set) — it's the only one that makes the trend a genuine per-test-speed signal.

## Links

Tracker #23 · timeline series #16 · `set_hash` fingerprint #24 · A/B macro #10

mparrett:
[...]

PR #270, opened 2026-09-05
docs: add production runtime and development benchmarking roadmap

nooga:
The compiler/VM audit identified concrete correctness, memory-lifetime, and performance gaps, but implementors need a sequenced plan and affordable ways to measure progress. This adds an extensive production-runtime roadmap with stable work-item IDs, prerequisites, design recommendations, acceptance criteria, dated audit evidence, reproduction recipes, and commit-pinned source references.

The benchmark design extends the existing tooling with bounded 20-second development checks, optional 120-second confirmation, cached binaries, paired samples, workload/result contracts, noise-aware regression decisions, and scheduled memory, latency, and peer comparisons. It addresses Go-specific representation and GC constraints and distinguishes proposed interfaces from commands that already exist.

Linked from the project README and performance documentation. Documentation only; no runtime or benchmark implementation changes.

Validation:
- `go build ./...` and `go test ./...` passed, including TestScripts.
- Checked local links, heading targets, pinned source paths/lines, JSON examples, and GitHub-rendered Markdown.
- Ran all five documented small JS reproductions against Paserati and Node; outcomes match the audit.
- Verified the documented short bench-ratchet capture command and a small same-binary paired capture. These validate examples/orchestration, not the proposed timing thresholds or statistical sensitivity.
- `git diff --check` passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,10 @@ Remaining gaps:
 
 See [docs/bucketlist.md](docs/bucketlist.md) for the exhaustive yet messy feature inventory.
 
+For an implementation plan covering compiler and VM correctness, Go memory
+management, interpreter performance, and bounded development benchmarks, see the
+[production runtime roadmap](docs/runtime-production-roadmap.md).
+
 ### Contributing
 
 Seriously, why would you want to contribute to this? _…But if you do, I'm both terrified and thrilled. PRs and issues are welcome._
--- a/docs/perf/README.md
+++ b/docs/perf/README.md
@@ -1,5 +1,11 @@
 # Perf
 
+The [production runtime roadmap](../runtime-production-roadmap.md#f-development-benchmarking-infrastructure)
+proposes an incremental extension of this tooling with bounded local checks,
+paired comparisons, workload contracts, and scheduled memory/latency runs. Its
+new commands and policies are design proposals; the commands below describe the
+existing implementation.
+
 `cmd/bench-ratchet` runs the Go benchmarks and normalizes each against a frozen
 calibration anchor (`BenchmarkRatchetAnchor`, a pure-CPU FMA loop in `pkg/vm`),
 so a benchmark reads as a multiple of the anchor's ns/op rather than an absolute
--- a/docs/runtime-production-roadmap.md
+++ b/docs/runtime-production-roadmap.md
@@ -1,0 +1,1386 @@
+# Production compiler and runtime roadmap
+
+Status: proposed implementation plan. Prepared 2026-09-05. This document changes
+no runtime behavior and implements none of the proposed tools or flags.
+
+Paserati's objective is a conformant, production-quality JavaScript/TypeScript
+runtime in Go, with excellent performance without a native JIT. The immediate
+strategy is to establish trustworthy correctness measurements, repair observable
+optimization and lifetime defects, and then reduce executed work, property-access
+cost, and memory traffic. Keep the direct TypeScript-to-bytecode pipeline.
+
+The September 2026 audit found substantial engineering and competitive performance
+against Goja. It did not establish production readiness or leadership among all
+interpreters. Go does not prevent most of the improvements below. It does impose
+constraints on value representation, collection isolation, and control over
+generated machine code. Those constraints should guide experiments, not substitute
+for measurements.
+
+This plan is intended to be split into small implementation PRs. Work-item IDs
+are stable references; their ordering expresses dependencies rather than dates.
+The acceptance criteria are proposed requirements. Performance thresholds and
+time budgets are initial policy choices to calibrate, not measured guarantees.
+
+## Navigation
+
+- [Evidence and scope](#evidence-and-scope)
+- [Architectural commitments](#architectural-commitments)
+- [Delivery sequence](#delivery-sequence)
+- [A: correctness and optimization contracts](#a-correctness-and-optimization-contracts)
+- [B: lifetime and representation](#b-lifetime-and-representation)
+- [C: compiler optimization](#c-compiler-optimization)
+- [D: property access and execution](#d-property-access-and-execution)
+- [E: production embedding](#e-production-embedding)
+- [F: development benchmarking infrastructure](#f-development-benchmarking-infrastructure)
+- [Implementation and release gates](#implementation-and-release-gates)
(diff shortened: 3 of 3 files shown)


mparrett:
A1 has a downstream consequence in the perf pipeline that I didn't see addressed, and the fix for it is already built and tested — it never went upstream.

## A1 moves `test262.total`

`test262.total` is the mean of per-test execution time over the tests that passed and did not time out. `cmd/bench-test262` states the intent plainly: *"restricting to the passing, non-timed-out set decouples speed from correctness."*

That decoupling holds only while the pass/fail classification is correct. A1 says it isn't — negative tests satisfied by wrong-phase errors or internal panics, and missing strict/module variants. So when A1 ships, the *membership of the timed population* changes, and the series moves for a reason that has nothing to do with how fast anything runs. It will read as a perf regression.

This is #26, still open. A1 makes it acute rather than theoretical, and it hits the one series the plan wants to use to qualify performance results.

## The reference set already exists

I built the fix for #26 on the fork on 2026-08-11 and never sent it up. `bench-test262` takes `-refset docs/perf/test262-refset.txt`, a pinned set of 40,141 tests that the mean is restricted to, so membership stops following the engine:

- The file verifies against its own header, because a file that merely exists proves nothing — the failure mode is a set that silently drifts from what it claims to be.
- Suite records carry their own share of the set. Sizing `test262.built-ins` against all 40,141 would make the shortfall arithmetic report every language test as a missing member.
- Seven tests cover it, including that a member dropping out is visible and that the set hash distinguishes equal-count different sets.

Happy to open that as a PR. It is worth merging *before* A1 rather than after, so the discontinuity has somewhere to be recorded instead of appearing as an unexplained step in the published history.

## Two numbers F1 can calibrate against

The plan proposes thresholds to calibrate rather than measured guarantees, so these may be useful.

**The layout floor.** A behaviour-neutral 27-line change to `pkg/vm` moves 10 of 42 benchmarks by more than 2%, maximum 4.2%, from code layout alone in the 261 KB `(*VM).run` (#52). Any threshold below that is unfalsifiable for `pkg/vm` edits, which bears directly on C and D.

**The minimum reducer is worse than the plan says.** F notes that a minimum "can conceal typical cost". Sharper: `PrototypeMethodAccess/WithDetailedStats/StringPrototypeMethod` has a sticky bimodal mode, ~300 ns and ~384 ns, that holds across all three samples of a `-count 3` capture — so min-of-N does not defend against it at all, and a single capture can be wrong by 27% in either direction, with the sign set by whichever side landed slow (#48). It is not per-process: the slow mode appears inside a single process, so fresh-process isolation does not remove it.

On F1's same-host paired comparison as the primary measurement, that is what I run for these, on a dedicated instance rather than a shared runner, interleaved and counterbalanced. The #143 measurement held anchor drift to 0.09% and resolves under 1%. The proposed `selfcheck` mode is the null control that produced the 4.2% floor above. Glad to put that behind F1, or to hand over the protocol if you'd rather it live entirely in-repo.

mparrett:
A concrete price tag for A1, from fixing an unrelated bug this afternoon.

`Promise.all` / `race` / `any` / `allSettled` / `try` called the species constructor instead of constructing it, so every combinator threw on a Promise subclass (#280). Fixing it moves reported `built-ins/Promise` conformance by **exactly zero** — 530 passing, 173 failing, before and after.

The reason is the success heuristic. The tests covering this are `built-ins/Promise/{all,race,any,allSettled,try}/ctx-ctor.js` and the `resolve-throws-iterator-return-*` family. The `TypeError` aborts the script before any assertion runs, and the aborted run scores as a pass — so the suite was already claiming those tests worked.

The probe is small enough to reuse as an acceptance check for A1. Append an unconditional throw to a test file:

```
throw new Error("REACHED_END");
```

On `main`, `ctx-ctor.js` with that line still reports pass, because execution never reaches it. Move the throw above the `Promise.all.call(...)` line and it correctly fails. A test whose appended throw does not change the verdict is not being executed to completion, whatever the suite reports.

Two consequences for the plan as written.

**The `-11` I cited on #48 and #50 was wrong**, and A1 is why. I read a set of Promise tests flipping to failing as regressions from a compiler change; they were false passes whose failure the change exposed. I have corrected both threads. Anyone using the current suite to qualify a change can make the same error in either direction, which is a sharper argument for A1 being a prerequisite than the fixture count alone.

**A1 will not only lower the reported percentage.** It will also make previously invisible repairs show up as gains. Both movements arrive in the same run, so the discontinuity cannot be read as "conformance dropped by N" — which is another reason to land the reference set (#26) before A1 rather than after, so the series has a stable membership to move against.

mparrett:
Correcting the "zero movement" datapoint above. It was measured on a binary built from `b8c17c8c`, before `502e0a9a` (fix exception loss in reentrant native calls) landed. On today's tip, `67d90d68`, reverting #280 takes `built-ins/Promise` from 530 / 173 to 519 / 184: the 11 tests in that cluster fail on `main` without it and pass with it.

So the swallowed-exception false pass was real, but it was a VM bug rather than a scoring heuristic: the `TypeError` thrown inside the reentrant native call was lost and the script ended clean. `502e0a9a` closed it, and its message lists those 11 as the real failures it exposed. The throw-append probe still works as an acceptance check for that class; the instance I offered is fixed.

The rest of the comment stands: `test262.total`'s membership follows the pass set, the reference set (#26) decouples it, and the layout floor and sticky mode numbers are unchanged.

nooga:
Thanks for the correction — good to have the "zero movement" number replaced
with the real one. Confirms `502e0a9a` closed the actual bug (exception loss
in reentrant native calls) rather than the runner's heuristic papering over it,
and that #280 does earn its 11 tests on current `main`.

The rest holds and I've folded it into the roadmap doc:

- Added #26 as an explicit dependency of A1, with the append-a-terminal-throw
  probe as its acceptance check. Please do open the PR for the reference-set
  fix — landing it before A1 so the `test262.total` step has somewhere to be
  recorded, rather than after, matches what you proposed.
- Added the #52 layout-floor (4.2%) and #48 sticky-bimodal numbers to F3 as
  calibration inputs for the initial timing budgets.

Nothing else needed on this thread — will track #26 and A1 sequencing through
their own issues from here.

PR #281, opened 2026-09-06
fix(test262): leave post-timeout cleanup to the test goroutine

mparrett:
After a timeout, `runSingleTest` waited 50 ms and then called `Cleanup()`, which sets `compiler`, `heapAlloc`, and the other fields to nil, while the test goroutine could still be inside `CompileProgram`. When `Compile` returned, the goroutine dereferenced the nil `heapAlloc` at `driver.go:383` and the run printed a compiler panic for a test that had only compiled slowly. Which tests hit it depends on machine load, so the count looks like a property of whatever compiler change is under test, which is how it got misread on #50.

This leaves cleanup to the goroutine's deferred call and drops the `sync.Once`. Nilling the fields early released nothing the goroutine's own stack still referenced.

## Verification

8 copies of a synthetic test with 16,000 function expressions at `-timeout 0.2s`: the previous runner prints 7 panic traces at that frame, this one prints none. The `language` suite failure set and the final memory footprint are unchanged.

### DIFF
--- a/cmd/paserati-test262/main.go
+++ b/cmd/paserati-test262/main.go
@@ -13,7 +13,6 @@ import (
 	"runtime/pprof"
 	"sort"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/nooga/paserati/pkg/builtins"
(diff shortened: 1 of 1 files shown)


nooga (approved):
Verified: Cleanup() nils p.heapAlloc while driver.go:383 dereferences it post-compile — the race this fixes is real. Combined merge builds and tests green.

PR #282, opened 2026-09-06
fix(compiler): reclaim block-scope registers in a fixed order

### LINKED ISSUE #50: compiler: register allocation is non-deterministic — one binary compiles the same source two different ways

mparrett:
Fixes #50.

`freeScopeRegisters` ranges over the scope's symbol map and frees each local's register as it goes. Go randomizes map iteration, `Free` pushes onto a LIFO free list, and the next allocation pops whichever register landed last, so one binary compiles one input several different ways. The variants are equivalent (a temporary lands in a different register), but they defeat bytecode hashing and make per-compile perf attribution noisy.

This collects the scope's freeable registers, sorts them, and unpins and frees in that order.

My earlier comments on #50 argued this change was unsafe because it moved test262 results and, on `0af3487d`, produced compiler panics. Neither holds. The moved tests were false passes (#280) and tests fixed upstream since, and the panics come from the test262 runner nilling the `Paserati` fields after a timeout while the test goroutine can still be inside `CompileProgram` (#281). Details on #50.

## Verification

- New `tests/compile_determinism_test.go` compiles a three-local block 200 times and requires one bytecode hash. It fails on main with three variants and passes here.
- test262 against `0af3487d`, same machine, sequential, `-timeout 0.2s`: `language` and `built-ins/{Promise,String,Object,Array,Function}` produce identical pass/fail sets before and after. `built-ins/RegExp` differs by two tests that sit on the timeout boundary and behave identically on both binaries when run alone.
- `go test ./...` green.

### DIFF
--- a/pkg/compiler/compiler.go
+++ b/pkg/compiler/compiler.go
@@ -1433,6 +1433,11 @@ func (c *Compiler) freeScopeRegisters(st *SymbolTable) {
 	if st == nil {
 		return
 	}
+	// Iterate in a fixed order. st.store is a map, so ranging it directly frees
+	// registers in Go's randomized map order; Free pushes onto a LIFO free list,
+	// so the next allocation pops whichever register happened to land last. That
+	// made one binary compile one input two different ways (nooga#50).
+	freeable := make([]Register, 0, len(st.store))
 	for _, sym := range st.store {
 		if sym.IsGlobal || sym.IsSpilled || sym.IsCallerLocal || sym.IsNamespaceProperty {
 			continue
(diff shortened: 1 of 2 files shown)


nooga (approved):
Verified against #50's history: the earlier -11 Promise / panic objections were confounders (test262 runner false-passes fixed by #280, timeout-cleanup race fixed by #281), not defects in this fix. Copied the new determinism test onto unpatched main and confirmed it fails there, passes here. Combined merge builds and tests green.

PR #286, opened 2026-09-06
fix(compiler): allocate the argument register of emitted error throws

mparrett:
The four emitters that throw a `ReferenceError` or `TypeError` from compiled code (TDZ access, unresolvable reference, strict undeclared assignment, const assignment) allocated the callee and result registers but loaded the message into `callee+1` without reserving it. `MaxRegs` never saw that register, so when the callee landed on the frame's last register the VM indexed one past its register file and reported an internal panic instead of throwing:

```
"use strict";
var n = 0;
var g = function* fn() { n++; fn = 1; };
g().next();
// [VM PANIC] recovered: runtime error: index out of range [6] with length 6
```

The temporaries from `n++` push the callee to the top of the frame. This is `language/expressions/generators/named-strict-error-reassign-fn-name-in-body.js`, and the same shape panics in a plain function once the frame is the right size. When `callee+1` fell inside the frame it aliased whatever lived there; I did not find a shape where that changes a result, since the throw ends the function, but the fix removes the possibility.

This folds the four into one helper that allocates callee and argument as a contiguous pair, which is the `OpCall` convention the code was relying on.

## Verification

- New `tests/scripts/strict_fn_name_assign_throw_registers.ts` panics on `main` and passes here.
- test262 on `33c00d2d`, `-timeout 0.2s`: `language` +1 (the test above), `built-ins/{Promise,String,Object,Array,Function}` unchanged.
- `go test ./...` green.

### DIFF
--- a/pkg/compiler/emit.go
+++ b/pkg/compiler/emit.go
@@ -943,122 +943,56 @@ func (c *Compiler) emitIteratorCleanupAbruptIfNotDone(iteratorReg Register, done
 	c.emitByte(byte(doneReg))
 }
 
-// emitTDZError emits code to throw a ReferenceError for accessing a variable
-// before it is initialized (Temporal Dead Zone violation in default parameters)
-func (c *Compiler) emitTDZError(hint Register, varName string, line int) {
-	// Allocate registers for function and result
-	funcReg := c.regAlloc.Alloc()
+// emitThrowNewError emits `throw new <ctorName>(msg)` for a global error
+// constructor. OpCall reads its argument from funcReg+1, so the callee and the
+// argument are allocated as one contiguous pair: writing to funcReg+1 without
+// reserving it left the frame one register short (the VM then indexed past the
+// register file) and could alias whatever local happened to live there.
+func (c *Compiler) emitThrowNewError(ctorName, msg string, line int) {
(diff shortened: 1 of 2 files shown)


nooga (approved):
Reviewed and independently verified:

- Reproduced the exact panic on `main` (`index out of range [6] with length 6`) using the PR's repro (`var g = function* fn() { n++; fn = 1; }` in strict mode, `g().next()`).
- Confirmed the PR branch throws `TypeError:1` cleanly instead of panicking.
- Root cause checks out: the four error-throw emitters wrote the message into `funcReg+1` without reserving it via the allocator, so `MaxRegs` undercounted the frame by one register — exactly the `OpCall` convention (`callee`, `args` at `callee+1..`) that's used consistently elsewhere in the compiler (e.g. `AllocContiguous` call sites in `compile_expression.go`).
- Fix consolidates the four near-duplicate emitters into one helper using `AllocContiguous(2)`, which is correct and reduces duplication.
- Full `go test ./...` green locally, plus the new regression test `tests/scripts/strict_fn_name_assign_throw_registers.ts` passes.
- CI green on all platforms, lint clean.

Minimal, correct, well-tested. LGTM.

PR #288, opened 2026-09-06
fix(parser): accept from, type, and satisfies as import binding names

mparrett:
`import from from "./m.js"` crashes the CLI:

```
panic: runtime error: invalid memory address or nil pointer dereference
  driver.(*Paserati).preloadNativeModules
```

`from` is lexed as a keyword, so the default-binding branch of `parseImportDeclaration` did not match and the specifier-list parser fell through every case and returned nil without recording an error. The resulting nil `*ImportDeclaration` went into the `Statement` interface as a non-nil value, nothing stopped the pipeline, and the module loader dereferenced it. `import type from`, `import * as from`, and `import { a as from }` failed the same way or with "Identifier expected". All four are legal: `from`, `type`, and `satisfies` are contextual keywords, not reserved words.

Three changes:

- Accept those tokens wherever an import binding is named (default binding, namespace alias, named alias). `type` still starts a type-only import unless it is directly followed by `from` or `,`.
- Record an error when the specifier list starts with a token none of the branches handle, so a failed import parse always surfaces.
- Unwrap the nil at the statement dispatcher, so a nil `*ImportDeclaration` can never reach later stages as a non-nil `Statement`.

This is the last of the four recovered VM panics the `language` suite printed on every run: the `source-phase-import` fixtures use `import source from ...` and `import from from ...` as default bindings, and the loader panicked in `extractImportSpecs` on the typed nil.

## Verification

- New `tests/scripts/import_binding_named_from.ts` imports `from`, `type`, `satisfies`, and a namespace from a helper module. It fails on `main` and passes here.
- test262 on `67d90d68`: `language/module-code/source-phase-import/import-source.js` goes from fail to pass; the rest of `language` and `built-ins/Promise` are unchanged once batch timeouts are rerun alone.
- `go test ./...` green.

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -3981,7 +3981,14 @@ func (p *Parser) curTokenIs(t lexer.TokenType) bool {
 // isContextualKeywordAsIdent checks if the current token is a contextual keyword
 // that can be used as an identifier (e.g., set, get, from, of, as, async, static, type)
 func (p *Parser) isContextualKeywordAsIdent() bool {
-	switch p.curToken.Type {
+	return isContextualKeywordType(p.curToken.Type)
+}
+
+// isContextualKeywordType reports whether t is a contextual keyword: reserved
+// in some grammar positions but a plain identifier everywhere else, including
+// as an import/export binding name.
+func isContextualKeywordType(t lexer.TokenType) bool {
+	switch t {
 	case lexer.SET, lexer.GET, lexer.FROM, lexer.OF, lexer.AS,
 		lexer.ASYNC, lexer.STATIC, lexer.TYPE, lexer.READONLY,
 		lexer.OVERRIDE, lexer.ABSTRACT, lexer.UNDEFINED, lexer.IS,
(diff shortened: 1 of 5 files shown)


nooga:
Pushed a follow-up commit that closes two gaps in the same fix:

- **Named-import specifier names still rejected `from`.** The specifier-list's imported-name check only accepted `IDENT`/`SATISFIES`, so while `from` worked as a default-import name (`import from from "m"`), it still failed as a named-import shorthand or alias source: `import { from } from "m"` and `import { from as x } from "m"` both errored with "Expected identifier, string literal, or 'default' in import specifier".

- **`isImportBindingName` only special-cased 3 of the 14 known contextual keywords.** The parser already has `isContextualKeywordAsIdent()`, used throughout for exactly this purpose, covering `from`, `type`, `satisfies`, `async`, `get`, `set`, `of`, `as`, `static`, `readonly`, `override`, `abstract`, `undefined`, `is`. The new `isImportBindingName` duplicated that list but narrowed it to just `from`/`type`/`satisfies`, so `import async from "m"`, `import get from "m"`, `import static from "m"`, etc. all still failed with "Expected '{', '*', or an identifier after 'import'" — the same class of bug, just not fully closed.

Fix: extracted `isContextualKeywordType(t)` out of `isContextualKeywordAsIdent()` as the single source of truth, and had `isImportBindingName` delegate to it instead of hand-picking a subset. Also updated the named-import specifier's imported-name check and its comma-continuation check to use `isImportBindingName`.

While in there, also fixed the invalid-alias-name error (`import { named as 123 }`): it was reported via `peekError`, which formats using `p.peekToken`, but the code had already advanced past `as` onto the offending token itself (`p.curToken`) — the error caret was landing one token too far right.

Added `import_binding_contextual_keywords.ts` (+ helper) covering the newly-fixed forms. Full smoke suite (`go test ./tests -run TestScripts`) and `go test ./...` are green.

nooga (approved):

PR #296, opened 2026-09-07
perf(vm): keep the tail-call [[HomeObject]] update out of the dispatch loop

### LINKED ISSUE #295: perf: 26a3bd68 costs 28% on Arith and 55% on MatrixMult on amd64 without running the new code
`26a3bd68` (preserve [[HomeObject]] across tail calls, #285) adds 14 lines inside `case OpTailCall` and `case OpTailCallMethod` of `(*VM).run`. On amd64 that grows `run`'s stack frame by four spill slots, and benchmarks that never reach those lines slow down well past the 3.7% layout floor measured for `./tests` on this box class.

## Measurement

Same box for every commit (`c7a.2xlarge`, EPYC 9R14, go1.26.0 linux/amd64), one compiled `./tests` binary per commit, pinned benchmark corpus overlaid, three interleaved counterbalanced rounds, median. Percent versus `a88125c2` (2026-08-31):

| benchmark | `0af3487d` | `f9569ee4` (parent) | `26a3bd68` | `67d90d68` |
|---|---|---|---|---|
| Arith | +1.1% | +1.0% | **+28.4%** | +27.9% |
| Factorial | -1.4% | -1.4% | **+11.3%** | +14.2% |
| MatrixMult | +2.4% | +2.3% | **+54.8%** | +5.8% |
| OctaneNavierStokes | -0.5% | -0.3% | **+35.8%** | +5.8% |
| OctaneRichards | +2.9% | +3.3% | +6.9% | +9.7% |
| OctaneDeltaBlue | +1.2% | +2.8% | +5.9% | +9.2% |
| Add | -0.3% | -0.2% | +0.6% | +1.5% |

The parent is flat; the step is the one commit. A five-round session over the wider range (`a88125c2`, `0af3487d`, `67d90d68`, plus a comment-only null control) agrees: the 45 commits up to `0af3487d` cost nothing the corpus can see and win 17% on Octane Splay, then the tip carries Arith +43%, Factorial +20%, NavierStokes +11%. Anchor drift across the session was 0.00%.

## Mechanism

`bench_arith.ts` is a loop of four arithmetic statements with no calls. `matrix_mult.ts` has no tail calls. `factorial.ts` recurses in operand position. None of them execute the new lines. What changed is the function they were added to, from `go tool nm` and `objdump` on the box's binaries:

| binary | `run()` bytes | frame |
|---|---|---|
| `f9569ee4` | 359,066 | `SUBQ $0x1f18, SP` |
| `26a3bd68` | 360,346 | `SUBQ $0x1f38, SP` |

`67d90d68` has a byte-identical `run()` and a different `pkg/compiler`; MatrixMult falls back to +6% there while Arith stays at +28%, so placement is involved as well as the frame. On an arm64 Mac the two commits are within noise (Arith -0.2%, Factorial -1.2%, n=8), so this is amd64 code generation of the dispatch loop.

## Remedy

Moving the block into a `//go:noinline` helper called from both cases returns the frame to the parent's size and makes `run()` 200 bytes smaller than before `26a3bd68`, with `tco_tail_call_super_homeobject.ts` still passing. I am measuring that on the same box now and will send it as a PR with the numbers.

`run()` is at a size where any edit inside it is a layout event. A frame-size check on `run()` in CI would catch this class for free; I'll send that as a separate PR unless you'd rather not gate on it.

mparrett:
Fixes #295.

`26a3bd68` inlined the [[HomeObject]] update into the `OpTailCall` and `OpTailCallMethod` cases of `(*VM).run`. On amd64 that grew `run`'s stack frame by four spill slots and the function by 1.3 KB, and benchmarks that never reach those lines paid for it. This moves the block into a `//go:noinline` helper called from both cases. Behaviour is unchanged; `tco_tail_call_super_homeobject.ts` still passes.

## Measurement

Same box (`c7a.2xlarge`, EPYC 9R14, go1.26.0 linux/amd64), one compiled `./tests` binary per commit, pinned corpus overlaid, four interleaved counterbalanced rounds, median. Percent versus the parent of `26a3bd68`:

| benchmark | `26a3bd68` | `0ec3c9a0` (main) | this PR | n |
|---|---|---|---|---|
| Arith | +37.2% | +26.3% | +0.2% | 12 |
| Factorial | +23.6% | +45.1% | -0.1% | 12 |
| MatrixMult | +6.4% | +5.1% | -0.2% | 12 |
| OctaneNavierStokes | +47.6% | +20.2% | -0.4% | 20 |
| OctaneRichards | +0.1% | +28.5% | +0.1% | 20 |
| OctaneDeltaBlue | +1.7% | +36.2% | -0.3% | 20 |
| OctaneRayTrace | +2.3% | +10.5% | -0.1% | 20 |
| SetIndex | +7.5% | +2.7% | +1.5% | 12 |
| Add | +1.8% | +0.1% | -1.1% | 12 |

`26a3bd68` and `main` share a byte-identical `run()` and land on different subsets, which is the placement half of the problem; with this change every benchmark is back on the parent.

From `go tool nm` and `objdump` on the same binaries:

| binary | `run()` bytes | frame |
|---|---|---|
| parent `f9569ee4` | 359,066 | `SUBQ $0x1f18, SP` |
| `26a3bd68` / `main` | 360,346 | `SUBQ $0x1f38, SP` |
| this PR | 358,874 | `SUBQ $0x1f18, SP` |

## Verification

- `go test ./...` green; `tco_tail_call_super_homeobject.ts` and the other `tco_*` scripts pass.
- The table above, four rounds on one box, with the parent as the null.

### DIFF
--- a/pkg/vm/call.go
+++ b/pkg/vm/call.go
@@ -25,6 +25,30 @@ func (e exceptionError) GetExceptionValue() Value {
 	return e.exception
 }
 
+// setTailCallHomeObject gives a tail-called callee its [[HomeObject]] for
+// super property access (see prepareCall below). Arrow functions read
+// their captured [[HomeObject]] (frame.closure.CapturedHomeObject) at use
+// time, so the frame's own field is left alone for them.
+//
+// Kept out of (*VM).run deliberately. Inlined in the OpTailCall and
+// OpTailCallMethod cases, these lines grew run's amd64 stack frame by four
+// spill slots and cost 28% on a call-free arithmetic loop that never reaches
+// them; the dispatch loop's register allocation is the thing being protected.
+//
+//go:noinline
+func setTailCallHomeObject(frame *CallFrame, calleeFunc *FunctionObject, thisVal Value) {
+	if calleeFunc.IsArrowFunction {
+		return
+	}
+	if calleeFunc.HomeObject.Type() != TypeUndefined && calleeFunc.HomeObject.Type() != TypeNull {
+		frame.homeObject = calleeFunc.HomeObject
+	} else if thisVal.Type() != TypeUndefined && thisVal.Type() != TypeNull {
+		frame.homeObject = thisVal
+	} else {
+		frame.homeObject = Undefined
+	}
+}
+
 // prepareCall sets up a function call and returns whether the interpreter should switch to the new frame.
 // For native functions, it executes immediately and returns false.
 // For closures/functions, it sets up the frame and returns true to switch context.
--- a/pkg/vm/vm.go
+++ b/pkg/vm/vm.go
@@ -1479,6 +1479,30 @@ func (vm *VM) InterpretWithCallerScope(chunk *Chunk, callerRegs []Value, callerT
 	return result, errs
 }
 
+// setTailCallHomeObject gives a tail-called callee its [[HomeObject]] for
+// super property access (see prepareCall in call.go). Arrow functions read
+// their captured [[HomeObject]] (frame.closure.CapturedHomeObject) at use
+// time, so the frame's own field is left alone for them.
+//
+// Kept out of (*VM).run deliberately. Inlined in the OpTailCall and
+// OpTailCallMethod cases, these lines grew run's amd64 stack frame by four
+// spill slots and cost 28% on a call-free arithmetic loop that never reaches
+// them; the dispatch loop's register allocation is the thing being protected.
+//
+//go:noinline
+func setTailCallHomeObject(frame *CallFrame, calleeFunc *FunctionObject, thisVal Value) {
+	if calleeFunc.IsArrowFunction {
+		return
+	}
+	if calleeFunc.HomeObject.Type() != TypeUndefined && calleeFunc.HomeObject.Type() != TypeNull {
+		frame.homeObject = calleeFunc.HomeObject
+	} else if thisVal.Type() != TypeUndefined && thisVal.Type() != TypeNull {
+		frame.homeObject = thisVal
+	} else {
+		frame.homeObject = Undefined
+	}
+}
+
 // run is the main execution loop.
 // It now returns the InterpretResult status AND the final script Value.
 func (vm *VM) run() (status InterpretResult, resultValue Value) {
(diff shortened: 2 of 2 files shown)


mparrett:
@nooga three things from checking this against `runtime-production-roadmap.md` and `AGENTS.md` before you review. Each has a default I'll take unless you say otherwise.

**1. D4 says no Go call per opcode without compelling measurements.** This adds one on the tail-call path, not per opcode, and the table above is the measurement. Two gaps: the corpus has no tail-call benchmark, so the helper's cost on tail-call-heavy code is asserted rather than measured, and the fix itself was measured on amd64 only (the regression is within noise on arm64, so the fix should be too, but that is inference). Default: merge as is, and I add a tail-recursive accumulator to the benchmark corpus and measure the helper against it in the next session. Alternative: hold the PR until that number exists.

**2. Placement.** `prepareCall` in `call.go` already makes the same three-way [[HomeObject]] decision for ordinary calls, and the original inline comment points there. The helper belongs beside it; sharing one function between both paths is a separate change, because adding a call inside `prepareCall` is its own perf question. Default: I move the helper into `call.go` in this PR, no behaviour or codegen change, and leave unification for a follow-up. Alternative: leave it next to `run()` where the reason for its existence is.

**3. The roadmap's layout floor is now stale.** It bounds `pkg/vm` thresholds with #52's 4.2%; #295 shows a 24-line edit moving Arith 28% without executing. The roadmap asks for its numbers to be updated with the implementing PR and replacement measurements. Default: I send a short docs PR after this merges, citing #295 and adding `run()`'s frame size as the cheap canary. Alternative: you fold it in, since the document is yours and three days old.

nooga:
Thanks @mparrett please take the default on all points.

mparrett:
Done for point 2: `306dd716` moves `setTailCallHomeObject` into `call.go` ahead of `prepareCall`. `run()` is byte-identical before and after the move (359,514 bytes, frame `0x1f58` on an amd64 cross-build here). Points 1 and 3 are queued: the tail-recursive accumulator goes into the corpus and gets measured in the next session, and the roadmap datapoint comes as a docs PR once this merges.

PR #307, opened 2026-09-07
fix(compiler,vm): fix array literal elision holes when mixed with a spread element

nooga:
## Summary

Follow-up to #305 (paserati#300). #305 fixed array literal elisions (`[1,,3]`) materializing as a present `undefined` instead of a real sparse hole — but only in `compileArrayLiteralSimple`. `compileArrayLiteralWithSpread` (used whenever the literal also contains a `SpreadElement`, e.g. `[1, , ...xs, , 2]`) was untouched by that fix and still turned those elisions into a real `undefined`:

```
const xs = [10, 20];
const mixed = [1, , ...xs, , 2];
1 in mixed, 4 in mixed  // was: true true — now: false false
```

## Root cause

`compileArrayLiteralWithSpread` builds the result incrementally. For each plain (non-spread) element it used to wrap the element in a synthetic 1-element array and `OpArraySpread` that into the destination. `OpArraySpread` reads its source through `vm.extractSpreadArguments`, which implements real iterator-protocol semantics — for an array that means `arr.Get(i)`, the same primitive that already resolves a `Hole` back into `Undefined`. So a `Hole` placed in the synthetic carrier array never survived the round trip.

## Fix

Added a new opcode, `OpArrayAppendRaw` (`Rx Ry`), that appends the raw value in `Ry` directly onto the array in `Rx` via a plain Go slice append (`ArrayObject.Append`) — no iterator, no property-set semantics. Unlike `OpArrayCopy` (also raw, but needs a compile-time-known destination offset — not available here since a prior spread can contribute a variable number of elements at runtime), `OpArrayAppendRaw` grows the array's length by one at whatever its current runtime length is.

`compileArrayLiteralWithSpread` now uses `OpArrayAppendRaw` for every plain element (elision or real value), and still uses `OpArraySpread` only for genuine `SpreadElement`s — which must keep resolving a hole in *their own* source array to `Undefined` per spec (intentionally unchanged; e.g. `[...[100,,300]]` still doesn't produce a hole at index 1 — that's a separate, already-tracked follow-up, not touched here). This also removes the previous double `OpMakeArray`+`OpArraySpread` indirection for plain elements in a spread-containing literal.

**Verified no behavior change from the iterator-protocol-override angle**: overriding `Array.prototype[Symbol.iterator]` before evaluating a literal with a spread does not affect the plain elements' values either before or after this change, since `extractSpreadArguments` uses a fast native-array path for arrays rather than consulting the prototype iterator. This fix only changes how plain elements reach the destination array (raw append vs. spread-through-iterator-of-a-synthetic-array), not how a real spread source is read.

## Tests

- Added `tests/scripts/issue300_array_holes_elision_with_spread.ts`: elision before/after/around a spread, multiple consecutive elisions, spread-only, leading/trailing elision alone, and confirming a real `undefined` next to a spread does not become a hole.
- Updated the stale "not yet fixed" comment in `tests/scripts/issue300_array_holes_literal_elision.ts`.
- `go test ./tests -run TestScripts`: all green.
- test262 `built-ins/Array` full suite (`-timeout 0.2s`): 0 diff, 0 regressions (3079 tests).
- test262 `language` full suite (`-timeout 0.2s`): 0 diff, 0 regressions (23523 tests).

## Base branch

This targets `fix-300-array-holes` (#305), not `main` — it depends on `ArrayLiteral.Elisions` and `compileArrayLiteralElement` introduced there and should merge after it.

### DIFF
--- a/pkg/compiler/compile_literal.go
+++ b/pkg/compiler/compile_literal.go
@@ -677,18 +677,21 @@ func (c *Compiler) compileArrayLiteralWithSpread(node *parser.ArrayLiteral, hint
 
 	// Process each element, adding to the result array
 	// Free registers eagerly to avoid exhaustion with large arrays
-	for _, elem := range node.Elements {
-		switch e := elem.(type) {
-		case *parser.SpreadElement:
+	for i, elem := range node.Elements {
+		if spreadElem, isSpread := elem.(*parser.SpreadElement); isSpread {
 			// Compile the spread expression (should be an array)
 			spreadReg := c.regAlloc.Alloc()
-			_, err := c.compileNode(e.Argument, spreadReg)
+			_, err := c.compileNode(spreadElem.Argument, spreadReg)
 			if err != nil {
 				c.regAlloc.Free(spreadReg)
 				return BadRegister, err
 			}
 
-			// Use OpArraySpread to append all elements from spreadReg to hint
+			// Use OpArraySpread to append all elements from spreadReg to
+			// hint - a real spread must read its source through the
+			// iterator protocol (which resolves a hole in the SOURCE array
+			// to a present Undefined, per spec), so this is deliberately
+			// NOT OpArrayAppendRaw.
 			c.emitOpCode(vm.OpArraySpread, line)
 			c.emitByte(byte(hint))      // DestReg: result array (modified in place)
 			c.emitByte(byte(spreadReg)) // SrcReg: array to spread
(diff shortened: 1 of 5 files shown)


nooga:
Superseded by #326 — this PR's base branch (fix-300-array-holes) was deleted when #305 merged, which auto-closed this PR. Rebased branch re-opened as #326 with no content changes.

PR #311, opened 2026-09-08
fix(builtins): audit pkg/builtins for len(args) vs IsUndefined() defect class

nooga:
## Summary

Full audit of `pkg/builtins/*.go`, following up on the two related bugs already in the stack: #310 (the compiler no longer pads a call site that omits a checker-known-optional argument with an explicit `undefined`, which was corrupting `arguments.length`) and #306 (`Array.prototype.join`'s separator, cherry-picked here as `a300dff7`).

Swept every `len(args) >= N`/`len(args) > N` site across all 43 `pkg/builtins/*.go` files (**340 raw grep hits**) for a native optional-parameter check that should instead (or also) check `!args[N-1].IsUndefined()`.

## The discriminator

In every case, the test is the relevant spec's **own wording** for that parameter, not just "is it optional":

- **"If X is undefined, default to Y"** (a VALUE check) — the `len(args)` check alone is wrong on its own, *independent of #310*, because it can't distinguish a real omission from an explicit `undefined` argument the caller wrote. Needs the guard.
- **"If X is not present, default to Y"** (a PRESENCE check) — #310 alone already makes both the omitted and explicit-`undefined` call shapes correct, because `arguments.length`/`len(args)` reflects the real call again. Adding an `IsUndefined()` guard here would be a **new** bug — it collapses a distinction the spec deliberately keeps.

## Confirmed genuine hits (value-based, fixed here)

- **`TextDecoder.prototype.decode`**'s `input` ([text_codec_init.go](pkg/builtins/text_codec_init.go)) — WebIDL optional `BufferSource` with no explicit default; WebIDL's ECMAScript binding treats omitted and explicit-`undefined` as equivalent. `decode(undefined)` was stringifying to the literal `"undefined"` instead of returning `""`.
- **`performance.getEntriesByName`**'s `type` ([performance_init.go](pkg/builtins/performance_init.go)) — same WebIDL rule, confirmed via a real-browser check. `getEntriesByName(name, undefined)` was filtering by the literal string `"undefined"` (matching nothing) instead of matching every entry type.
- **`Number.parseInt`**'s `radix` ([number_init.go](pkg/builtins/number_init.go)) — not really a presence/value question; `ToInt32(radix)` spec-mandates the same outcome (R=0, "unspecified", default base 10) for `NaN`, `+Infinity`, and `-Infinity` alike. The bug was `int(args[1].ToFloat())` on a non-finite float — Go-implementation-defined, happened to yield 0 on this platform for NaN specifically (masking the bug), but wasn't portable and didn't cover `Infinity`/`-Infinity` at all. Routed through the existing (previously dead-code) `toInt32()` helper in `math_init.go`. Verified `parseInt("10", Infinity/-Infinity/NaN/undefined)` all now return `10`, matching Node.js exactly — including the `1e20` edge case, which correctly still returns `NaN` in both paserati and Node.

## Examined and confirmed to need NO change (presence-based)

Verified side-by-side (omitted vs. explicit `undefined`) against this branch, matching Node.js:

- `Array.prototype.reduce`/`reduceRight` and `TypedArray.prototype.reduce`/`reduceRight`'s `initialValue` — explicit `undefined` seeds the accumulator as `undefined` (→ `NaN` on arithmetic); omitted seeds from the first element.
- `Array.prototype.splice`/`toSpliced`'s `deleteCount` — present-as-undefined runs `ToIntegerOrInfinity(undefined) = 0` (deletes nothing); not-present deletes to the end. Verified `arr.splice(1, undefined, "x")` still correctly deletes zero elements.
- Every `Date.prototype.set*`/`setUTC*` trailing time/date component — present-as-undefined runs `ToNumber(undefined) = NaN` (invalidates the date); not-present keeps the current value. Verified `setMonth(5, undefined)` still correctly produces an Invalid Date.
- `Iterator.prototype.reduce` (same rule as `Array.prototype.reduce`).
- Every other file (DataView/ArrayBuffer/SharedArrayBuffer/Atomics/BigInt, String/RegExp/TextEncoder, Object/Map/Set/WeakMap/WeakSet/JSON/Reflect/globals/console, Promise/Error/generators/Function/DisposableStack/FinalizationRegistry, fetch/Blob/ReadableStream/FormData) — each hit was already guarded, a genuinely required parameter, count-based dispatch the spec itself describes as varying by argument count, or a coded fallback that already produces the same result for `undefined` as for omission.

## Explicitly NOT touched here — spawned as separate follow-up tasks

- `Blob.prototype.slice`'s omitted `contentType` defaulting to the parent blob's own type instead of the File API spec's `""` — a wrong-default-value bug unrelated to this audit's mechanism.
- `typedarray_common.go`'s `SetupTypedArrayPrototype` — dead code (zero callers) carrying its own stale `reduce`/`reduceRight` copies with no runtime effect.
- `AbortSignal.abort`/`AbortController.abort`'s `reason` parameter — same `len(args)`-only pattern, but currently unreachable because its TS signature declares zero parameters; flagged so completing that signature later doesn't silently reintroduce this bug class.

## Tests

Added [optional_param_undefined_vs_omitted.ts](tests/scripts/optional_param_undefined_vs_omitted.ts), run in **default type-checked mode** specifically because that's what's being verified. Asserts both call shapes for both categories — the three value-based fixes (where both shapes must agree) and reduce/splice/a Date setter (where they must deliberately differ) — so a future regression in either direction fails loudly.

## Testing

- `go test ./tests -run TestScripts`: all green
- test262 `built-ins/Number`, `built-ins/Array`, `built-ins/Date` (`-timeout 0.2s`): **0 diff, 0 regressions**, isolated specifically to this commit's 3 fixes (compared against the branch state immediately before them) — expected, since TextDecoder/Performance aren't test262-covered and the `parseInt` fix only changes behavior for non-finite radix inputs test262 doesn't exercise this way.

## Base branch

This stacks on **#310** and **#306** (cherry-picked as `a300dff7`) and must merge after both — the new test's presence-based assertions (`reduce`/`splice`/`setMonth`) only pass because #310's argument-padding removal is present. **Merging this before #310 would make the smoke suite red.**

### DIFF
--- a/pkg/builtins/array_init.go
+++ b/pkg/builtins/array_init.go
@@ -677,8 +677,21 @@ func (a *ArrayInitializer) InitRuntime(ctx *RuntimeContext) error {
 		if err != nil {
 			return vm.Undefined, err
 		}
+		// Per ECMA-262 23.1.3.16 Array.prototype.join step 3: "If separator
+		// is undefined, let sep be the single-element String ','" - an
+		// explicit `undefined` argument must default to "," exactly like an
+		// omitted one, not stringify to the literal "undefined". This also
+		// matters when type checking is enabled: the compiler pads a call
+		// site to a native function's declared (checker-known) arity with
+		// explicit `undefined` arguments for any parameter the checker
+		// considers optional-but-unprovided (compileArgumentsWithOptionalHandling,
+		// pkg/compiler/compiler.go), so `arr.join()` and `arr.join(undefined)`
+		// are indistinguishable from this function's point of view - only a
+		// `len(args) >= 1` check (as opposed to checking the value itself)
+		// would wrongly treat the padded call the same as an explicit
+		// separator.
 		separator := ","
-		if len(args) >= 1 {
+		if len(args) >= 1 && !args[0].IsUndefined() {
 			separator = args[0].ToString()
 		}
 		if length == 0 {
--- a/pkg/builtins/number_init.go
+++ b/pkg/builtins/number_init.go
@@ -765,8 +765,23 @@ func (n *NumberInitializer) InitRuntime(ctx *RuntimeContext) error {
 
 		str := args[0].ToString()
 		radix := 10
+		// Per spec (18.2.5 parseInt, which Number.parseInt is required to be
+		// the same function object as), radix goes through ToInt32 - which
+		// spec-mandates +0 for NaN and either Infinity (see toInt32/toUint32
+		// in math_init.go: NaN/±Infinity/±0 all map to 0). A radix of 0 is
+		// then treated as "unspecified" and defaults to 10 below. The
+		// previous code did an unconditional `int(args[1].ToFloat())`, which
+		// for a non-finite radix (an omitted argument padded to `undefined`
+		// by the compiler before paserati#310, or a real `Infinity`/NaN the
+		// caller wrote) relies on Go's implementation-defined float64->int
+		// conversion of a non-finite value - happens to yield 0 on this
+		// platform (coincidentally falling into the `radix == 0` default
+		// below), but is not portable and doesn't cover every non-finite
+		// input (Infinity, -Infinity, 1e20 all reproduced the same bug
+		// empirically, not just NaN). Route through the real ToInt32
+		// instead.
 		if len(args) > 1 {
-			radix = int(args[1].ToFloat())
+			radix = int(toInt32(args[1].ToFloat()))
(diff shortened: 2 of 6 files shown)


nooga:
Superseded by #327 — this PR's base branch (fix-optional-param-arguments-length) was deleted when #310 merged, which auto-closed this PR. Rebased branch re-opened as #327 with no content changes.

Issue #402, opened 2026-09-11
Spike: B1 frame/register lifetime repair (+ B4 segmented stack follow-up)


nooga:
Tracks roadmap work items [B1](https://github.com/nooga/paserati/blob/main/docs/runtime-production-roadmap.md#b1-release-dead-register-and-frame-references) (release dead register/frame references) and [B4](https://github.com/nooga/paserati/blob/main/docs/runtime-production-roadmap.md#b4-replace-eager-maximum-register-allocation-with-stable-segments) (replace eager `256*4096` register allocation with stable segments), per `docs/runtime-production-roadmap.md`.

Concrete bugs found in this area, all fixed under this spike:

- #399 - register-stack leak (~1M unaccounted registers) surfacing as a crash in the fixed-size `256*4096` register file - fixed, confirmed against the real repro
- #400 - bound-constructor `new` frame push missing `newFrame.allocatedRegSize` - fixed
- #401 - `ActionReturn` pending-action resumption is a no-op stub, discards return value, never pops frame - fixed
- #404 (found while landing the invariant checker below) - derived-constructor-return TypeError/ReferenceError over-reclaims registers when caught above the immediate caller - fixed via `9ab46c89`

Phase 1 landed in #403 (closes #399, #400, #401 on merge) and #405 (invariant checker).

## Phase 1 (B1 - correctness, do first)

- [x] Fix #400 (concrete missing-field bug) - #403
- [x] Implement #401 (`ActionReturn` resumption), mirroring `OpHandlePending`'s existing correct arm rather than a from-scratch factor-out (left a `TODO(#401)` to unify the three copies later) - #403
- [x] Audit all frame-exit paths for register/frame cleanup - found and fixed the same `function.RegisterSize` vs `frame.allocatedRegSize` bug in `OpReturnUndefined`, `OpReturnFinally`, `OpHandlePending`'s `ActionReturn` arm, and `resumeGeneratorWithException`'s stale-frame cleanup (#399's actual root cause, confirmed against the real tsc.js/lib.dom.d.ts repro on the noderati side). Tail-call replacement, cancellation, generator close, and Reset were spot-checked and look consistent, not exhaustively fuzzed - #403
- [x] Add the trigger-relative push/pop invariant check #399 suggests - #405 (`vm.checkRegWindowRelease`, gated behind `vm.StrictRegWindowChecks`; landing it unconditionally found #404, filed separately rather than blocking this)

[...]

Issue #401, opened 2026-09-11
vm: ActionReturn pending-action resumption is a TODO stub - silently discards the return value and never pops the frame


nooga:
## Summary

`pkg/vm/vm.go`'s pending-action resumption logic (the `if vm.finallyDepth == 0 && vm.pendingAction != ActionNone` block that runs after a `finally` handler completes) has a `case ActionReturn:` arm that is a literal no-op stub:

```go
case ActionReturn:
    // Resume the return with saved value
    vm.pendingAction = ActionNone
    _ = vm.pendingValue // TODO: Implement return logic
    vm.pendingValue = Undefined
    continue
```

Compare with the sibling `case ActionThrow:` arm immediately above it, which fully re-drives the exception through `vm.throwException`. `ActionReturn` instead: discards the saved return value entirely, never pops the returning function's frame, and `continue`s the dispatch loop from wherever `ip` currently sits (right after the finally block) - as if the `return` statement that set `pendingAction = ActionReturn` in the first place (see the two call sites in `OpReturn`/`OpReturnUndefined` that do `vm.pendingAction = ActionReturn; vm.pendingValue = result; ip = handler.HandlerPC; continue`) had never happened.

## What this means in practice, if this path is ever actually reached

- The function's real return value is silently lost (replaced with whatever the function naturally produces by falling through past the `finally` block instead of returning at that point).
- Control flow does not actually leave the function at the `return` statement - execution continues into whatever code textually follows the `try/finally`, which real JS semantics say should never run.

## Not (yet) confirmed to be reachable in practice

Instrumented directly while investigating #399 (a real register-stack leak in the same file): a print inside this exact `case ActionReturn:` arm fired **zero** times across a large, real-world workload (running `typescript@5.8.3`'s `tsc.js` binder over `lib.dom.d.ts`), including plain and loop-nested `try { return x; } finally { ... }` shapes tested directly, both of which executed correctly. This suggests the compiler currently inlines the finally block at those exit edges rather than ever routing through this runtime resumption path - so this dead-looking code may currently be **unreachable** with the current compiler's output. Filing regardless since:
1. It's clearly an intentionally-incomplete stub (the `TODO` comment says so).
2. If it *is* unreachable today, either the compiler should be relied upon never to emit code that reaches it (worth a comment saying so, if confirmed), or it's dead code guarding against a future compiler change that would need this to actually work.
[...]

Issue #400, opened 2026-09-11
vm: new on a bound constructor's frame push is missing newFrame.allocatedRegSize (register leak on return)


nooga:
## Summary

`pkg/vm/vm.go`'s bound-constructor `new` path (the branch handling `new boundFn()` where `boundFn = originalFn.bind(...)`, in the big `switch` that handles constructor calls) pushes a new call frame without setting `newFrame.allocatedRegSize`.

Every other frame-push site that carves a window out of `vm.registerStack` sets this field immediately:

```
pkg/vm/call.go:416:        newFrame.allocatedRegSize = requiredRegs // Track actual allocation for proper cleanup
pkg/vm/op_spreadnew.go:232: newFrame.allocatedRegSize = requiredRegs // Track actual allocation for proper cleanup
pkg/vm/op_spreadnew.go:355: newFrame.allocatedRegSize = requiredRegs // Track actual allocation for proper cleanup
pkg/vm/vm.go:11987:         newFrame.allocatedRegSize = requiredRegs // Track actual allocation for proper cleanup
pkg/vm/vm.go:12222:         newFrame.allocatedRegSize = requiredRegs // Track actual allocation for proper cleanup
```

The bound-constructor site (currently around `pkg/vm/vm.go:12710`, inside the `case TypeClosure:` branch under the bound-function constructor handling) does:

```go
newFrame.registers = vm.registerStack[vm.nextRegSlot : vm.nextRegSlot+requiredRegs]
vm.nextRegSlot += requiredRegs

// Copy combined args to registers
for i := 0; i < len(newFrame.registers); i++ {
```

...with no `newFrame.allocatedRegSize = requiredRegs` in between. Since `vm.frames` is a preallocated, reused array, the frame slot's `allocatedRegSize` field retains whatever stale value was left over from that slot's *previous* occupant. `OpReturn` later reclaims register space via `returningFrameRegSize := frame.allocatedRegSize` (`pkg/vm/vm.go:6384`) - so on return from this call, `vm.nextRegSlot` is decremented by the stale value instead of `requiredRegs`, permanently leaking the difference (or, if the stale value happens to be larger, over-reclaiming) every time a bound function is used as a constructor.

## Repro shape

```js
function Base(x) { this.x = x; }
const Bound = Base.bind(null, 1);
for (let i = 0; i < N; i++) {
  new Bound();
}
```

[...]

PR #403, opened 2026-09-11
fix(vm): repair register/frame lifetime bugs behind #399/#400/#401 (B1)

### LINKED ISSUE #399: vm: real register-stack leak surfaces running tsc.js's binder over lib.dom.d.ts (Register stack overflow, ~1M registers unaccounted for)
## Summary

Running the real, unmodified `typescript@5.8.3` npm package's `tsc.js` under noderati (this repo's sibling Node-compat project, https://github.com/nooga/noderati) crashes with a `Register stack overflow` inside the VM's own register-stack accounting — **not** a legitimate deep-recursion overflow. At the crash point, only 71 real (live) call frames exist, and the sum of their own `RegisterSize` is 2148, but `vm.nextRegSlot` is already at 1,048,575 of a 1,048,576-slot register file (`RegFileSize(256) * MaxFrames(4096)`). ~1,046,000 registers are unaccounted for by any currently-live frame — a real register-stack leak, not a depth or sizing problem.

## Repro

```
git clone https://github.com/nooga/noderati
cd noderati && go build -o /tmp/noderati_bin ./cmd/noderati
cd examples
cat > /tmp/hello.ts <<'EOF'
function greet(name: string): string { return "hello " + name; }
console.log(greet("x"));
EOF
cat > /tmp/tsconfig.dom.json <<'EOF'
{
  "compilerOptions": { "target": "es2018", "module": "commonjs", "noEmit": true, "lib": ["es2018", "dom"] },
  "files": ["/tmp/hello.ts"]
}
EOF
/tmp/noderati_bin node_modules/typescript/lib/tsc.js -p /tmp/tsconfig.dom.json
```

Crashes every time with:

```
=== VM Stack (register overflow) ===
    at ... (bind/bindContainer/bindChildren/bindEachChild/forEach chain, ~71 frames)
    at bindSourceFile2
    at bindSourceFile
    at initializeTypeChecker
    at createTypeChecker
    ...
```

## The discriminating variable: `lib` size, isolated to `dom`

- `lib: ["es5"]` on the same trivial `hello.ts` → **exit 0, clean.**
- `lib: ["es2018"]` (no dom) → **exit 0, clean.**
- `lib: ["es2018", "dom"]` → **crashes**, same trace shape every time.
- Full default lib set (no explicit `lib`, which for `es2018` target implicitly includes DOM) → **crashes identically.**

So this isn't about recursion depth in the abstract — `bindSourceFile`'s real binder pass over `lib.es5.d.ts`/`lib.es2018.d.ts` alone completes fine. Something about processing `lib.dom.d.ts` specifically causes register-stack space to accumulate far beyond what any live frame needs.

## What the numbers show

At the crash, instrumented (`vm.frameCount`, `vm.nextRegSlot`, and a walk of `vm.frames[0:frameCount]` summing each live frame's `closure.Fn.RegisterSize`):

```
frameCount=71 nextRegSlot=1048575 regStackLen=1048576
sum of live frames' RegisterSize = 2148
```

A ~1,046,427-register gap between what's live and what's allocated. This is consistent with many already-*returned* calls each failing to give back their register window on return — a leak proportional to total binder work done, not to actual live call-stack depth. That also explains why `dom` (a much larger declaration surface than `es5`/`es2018` alone) is what tips it over: more total binder invocations, more accumulated leak, same live-depth ballpark (~15-90 frames throughout).

## Growth signature (best lead for finding the exact call site)

Sampling `vm.nextRegSlot` every time `vm.frameCount` returns to a shallow value (≤20) and printing only new highs showed, later in the same run, a long, perfectly monotonic **`+3` per sample, frameCount pinned at exactly 20**, sustained over ~1000 consecutive samples (`51000, 51003, 51006, 51009, ...`). That's the shape of one specific, small (`RegisterSize == 3`?) function/call-path whose net effect on `nextRegSlot` across a full call+return round trip is `+3` instead of `0` — i.e., a fixed per-call leak, not variable-size legitimate work.

**Caveat on reproducing that specific window**: a follow-up run gated on the exact numeric window (`nextRegSlot` between two absolute values, `frameCount` between 17-21) that had shown the pattern in an earlier build produced zero hits, even though nothing about the leak mechanism itself should have changed between builds. This is likely because tsc.js's own execution touches JS engine object/Map iteration order in enough places that the *exact* numeric trajectory of `nextRegSlot` isn't bit-reproducible run to run, even though the qualitative bug (and the final crash) reproduces every single time. **Any future instrumentation aimed at finding the exact opcode should be trigger-relative** (e.g., record `{funcName, allocatedRegSize}` at every push when frameCount is in some shallow band, and assert at the matching pop that the reclaimed amount equals what was recorded — flagging a mismatch immediately with full context) rather than keying off absolute `nextRegSlot` value windows.

## What this round ruled out (so the next one doesn't have to re-derive it)

Four plausible mechanisms were instrumented and checked directly against this exact repro; none of them fired anywhere near enough (or at all) to explain the leak:

1. **`executeUserFunctionSafe`/`executeUserFunctionWithNewTarget`'s documented native-boundary exception-unwind leak** (`truncateFramesTo` in `pkg/vm/vm_init.go`, cross-referenced with issue #61's discussion) — real, but only ever accounted for **269 total leaked registers** across the whole DOM-lib run (2 events). Not the primary driver.
2. **The `ActionReturn` "pending action" resumption path is a literal no-op stub** (`pkg/vm/vm.go`, the `case ActionReturn:` arm under `if vm.finallyDepth == 0 && vm.pendingAction != ActionNone`, comment reads `// TODO: Implement return logic`, discards `vm.pendingValue` and doesn't pop the frame). This is a real, separate, confirmed bug — return-in-finally *should* route through it in some compiled shapes — but a direct instrumented count showed it fires **0 times** in this entire workload. Simple and loop-nested `try { return x } finally { ... }` repros both worked correctly (compiler apparently inlines the finally at those exit edges rather than using this runtime path), so this bug exists but isn't this leak. Worth its own issue/fix regardless.
3. **A `new` on a bound constructor (`new (Fn.bind(...))()`) is missing `newFrame.allocatedRegSize = requiredRegs`** at its frame-push site (`pkg/vm/vm.go`, ~line 12710 in the `OpNew`-family bound-function-construction branch) — the only one of the six `newFrame.registers = vm.registerStack[...]` push sites across `call.go`/`op_spreadnew.go`/`vm.go` missing that assignment. Confirmed via inspection (grep for `allocatedRegSize` next to each of the six sites) and via a print showing the stale/uninitialized value at that exact line. Real bug (would leak `RegisterSize` minus whatever stale value happened to be left over from that frame slot's last use, every time this exact JS pattern executes), but instrumented and confirmed to fire **0 times** in this workload — not this bug's cause. Worth fixing on its own merits.
4. **`OpTailCall`'s "never shrink" register-window policy** (`pkg/vm/vm.go`, `// Note: We do NOT shrink! Bytecode may reference registers beyond RegisterSize`) — inspected directly; appears to be a deliberate, internally-consistent design (a tail-call chain's peak register usage is reclaimed once, in full, when the reused frame finally returns via the ordinary `OpReturn`/`OpReturnUndefined` path using `frame.allocatedRegSize`). Not ruled out with the same rigor as 1-3 (not instrumented directly), but nothing in a code read suggests an accounting bug here.

## Environment

- paserati: current `main` at the time of this report.
- noderati: current `main`.
- Repro is 100% deterministic in outcome (always crashes on `dom`, always clean on `es5`/`es2018`-only), even though the *exact* numeric growth trajectory sampled via instrumentation was not bit-reproducible run to run (see caveat above).

## Suggested next step

Add a trigger-relative invariant check (as described above) scoped to the normal `call.go` push / `OpReturn`+`OpReturnUndefined` pop pair first (the most common path), then expand to `op_spreadnew.go` and the two `OpCallMethod`-shaped fast-call sites in `vm.go` if that doesn't find it. The `+3`, flat-`frameCount=20` signature strongly suggests a single small, frequently-called function (plausibly a TS binder helper called once per DOM-lib declaration) with a fixed-size per-call leak.

nooga:
Phase 1 of #402 (B1 frame/register lifetime spike).

## What this fixes

- **#400** - `new` on a bound constructor pushed its frame without setting `newFrame.allocatedRegSize`, so `OpReturn` later reclaimed a stale leftover value instead of the real allocation.
- **#399** - `OpReturnUndefined`, `OpReturnFinally`, and `OpHandlePending`'s `ActionReturn` arm all reclaimed `function.RegisterSize` (the currently executing function's own size) instead of `frame.allocatedRegSize` (the frame's actual, possibly TCO-expanded allocation - TCO never shrinks a frame). In a tail-call chain where an earlier link needed more registers than the one that ultimately returns through one of these three paths, this permanently leaked the difference on every call - matching #399's "+3 per call, frameCount pinned" signature exactly. `OpReturn` itself already had this right. Also fixed the same pattern in `resumeGeneratorWithException`'s stale-frame cleanup. Confirmed against the real noderati/tsc.js `lib.dom.d.ts` repro - see #399.
- **#401** - `ActionReturn`'s pending-action resumption was a no-op stub that discarded the return value and never popped the frame. Implemented it by mirroring `OpHandlePending`'s already-correct `ActionReturn` arm (constructor semantics, sentinel/direct/normal-caller resume all preserved). Left a `TODO(#401)` noting the three near-duplicate copies of this pop-and-resume logic should eventually be factored into one shared helper.

## Testing

- Added two smoke tests that reproduce a register-stack overflow before the fix (confirmed against a locally reverted build) and pass cleanly after:
  - `tests/scripts/tco_regsize_leak_on_implicit_return.ts`
  - `tests/scripts/bound_constructor_regsize_leak.ts`
- Full `TestScripts` smoke suite and `go test ./...` pass with no regressions.
- `#401`'s fix has no dedicated repro - its own investigation found the path currently unreachable by the compiler's output - but the full try/finally/generator test suite passes unchanged.
- The original tsc.js/`lib.dom.d.ts` crash from #399 was re-run against this branch on the noderati side and confirmed fixed (my own sibling-checkout attempt at reproducing it didn't hit the same code path either way - see the discussion on #399).

## Not in this PR

- The trigger-relative push/pop invariant checker #399 suggests, and a full audit sweep beyond the return/exception-unwind paths (tail-call replacement, cancellation, generator close, Reset were spot-checked, not exhaustively fuzzed).
- B4 (segmented register stack) - deferred until this lands, per the roadmap's own sequencing ("B1 must be correct first").

Closes #399, #400, #401.
Related to #402.

### DIFF
--- a/pkg/vm/vm.go
+++ b/pkg/vm/vm.go
@@ -6678,7 +6678,11 @@ startExecution:
 			if debugVM {
 				fmt.Printf("[DBG] Popping frame...\n")
 			}
-			returningFrameRegSize := function.RegisterSize
+			// Use allocatedRegSize (not function.RegisterSize) which tracks actual
+			// allocation - may differ from the current function's own RegisterSize
+			// because of TCO expansion by an earlier link in a tail-call chain
+			// (see the matching comment on OpReturn above, and #399/B1).
+			returningFrameRegSize := frame.allocatedRegSize
 			callerTargetRegister := frame.targetRegister
 			isConstructor := frame.isConstructorCall
 			constructorThisValue := frame.thisValue
@@ -12708,6 +12712,7 @@ startExecution:
 					newFrame.args = finalArgs
 					newFrame.argumentsObject = Undefined
 					newFrame.registers = vm.registerStack[vm.nextRegSlot : vm.nextRegSlot+requiredRegs]
+					newFrame.allocatedRegSize = requiredRegs // Track actual allocation for proper cleanup
 					vm.nextRegSlot += requiredRegs
 
 					// Copy combined args to registers
@@ -15688,7 +15693,11 @@ startExecution:
 			}
 
 			// Pop the current frame (same logic as OpReturn)
-			returningFrameRegSize := function.RegisterSize
+			// Use allocatedRegSize (not function.RegisterSize): it tracks actual
+			// allocation, which may exceed the current function's own RegisterSize
+			// because of TCO expansion by an earlier link in a tail-call chain
+			// (see the matching comment on OpReturn/OpReturnUndefined, and #399/B1).
+			returningFrameRegSize := frame.allocatedRegSize
 			callerTargetRegister := frame.targetRegister
 			isConstructor := frame.isConstructorCall
 			constructorThisValue := frame.thisValue
@@ -15883,7 +15892,9 @@ startExecution:
 				}
 
 				// Pop the current frame
-				returningFrameRegSize := function.RegisterSize
+				// Use allocatedRegSize (not function.RegisterSize): see the matching
+				// comment on OpReturn/OpReturnUndefined/OpReturnFinally (#399/B1).
+				returningFrameRegSize := frame.allocatedRegSize
 				callerTargetRegister := frame.targetRegister
 				isConstructor := frame.isConstructorCall
 				constructorThisValue := frame.thisValue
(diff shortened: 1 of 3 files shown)


nooga:
Re-ran #399's actual repro (real, unmodified `typescript@5.8.3`'s `tsc.js` under noderati, `lib: ["es2018", "dom"]` against a trivial `hello.ts`) against this branch (`e8ea9e8b`), from the noderati side:

- The previously-100%-reproducible `Register stack overflow` crash is **gone** - clean exit 0.
- Also re-ran the default-lib case (no explicit `lib`, which implicitly includes DOM for an `es2018` target) - clean.
- The two known-good fixtures (`es5`-only, `es2018`-only-no-dom) still pass, as does `--version`/`--help`.
- Re-ran a real two-file TS program (interface/class/imports, `strict: true`, real emit) through `tsc.js` and executed the emitted JS - output still matches real Node byte-for-byte.
- noderati's own `internal/host` suite: green (`go test ./internal/host/... -skip TestEventsAddAbortListener`, ~9.6s).
- This branch's own `go build ./...`/`go vet ./...`/`go test ./...` (including the two new smoke tests): all green.

Confirms the `function.RegisterSize` vs `frame.allocatedRegSize` fix in `OpReturnUndefined`/`OpReturnFinally`/`ActionReturn` was in fact #399's root cause - the tsc.js/lib.dom.d.ts workload is exactly the kind of tail-call-heavy binder code that would hit it. Good to close #399 once this merges.

PR #435, opened 2026-09-12
test: fix stale nested-if depth in new_function_deeply_nested_clean_error.ts (main is currently red)

nooga:
## Summary

**\`main\` currently fails \`go test ./tests -run TestScripts\`.** This test asserts that a 150-level nested-if \`new Function()\` body throws a catchable \`SyntaxError\`. That was true when #430 added it, but #431 (the nested-if/ternary register-leak fix, merged separately) raised the register-exhaustion ceiling from ~118 to over 1000 nested levels — so depth 150 no longer errors at all once both land on \`main\` together. Each PR was correct and green individually; combined, they leave this test stale.

Bumped the depth to 2000, which reliably hits the *next*, unrelated, pre-existing \`"function too large: ... exceeds the 16-bit branch limit"\` bytecode-size error either way — keeping the test's actual assertion (a deeply-nested \`new Function()\` body fails as a catchable \`SyntaxError\`, never crashes) meaningful regardless of the register ceiling.

No production code changes — test-only.

## Verified
- \`go test ./tests -run TestScripts\` and \`go test ./...\` both green on this branch.

### DIFF
--- a/tests/scripts/new_function_deeply_nested_clean_error.ts
+++ b/tests/scripts/new_function_deeply_nested_clean_error.ts
@@ -1,18 +1,29 @@
 // expect: true
 // paserati#426: a `new Function(...)` body large/deeply-nested enough to hit
-// the compiler's internal register-allocator limit (this synthetic shape -
-// 150 sequential nested `if`s - mirrors real generated code from JSON-Schema
-// validators, bundled SDK clients, etc.) must surface as a normal, catchable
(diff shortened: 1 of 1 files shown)


nooga:
Closing as redundant — #434 (merged) already included this same depth-150→2000 fix to this test file, so `main` already has it. This PR's diff no longer applies cleanly on top of that (hence the conflict).

Issue #117, opened 2026-08-30
compiler/vm: a dependency module's own function body can't read its own module's top-level const/let after import — ReferenceError


nooga:
## Bug

A function declared in a **non-entry** module, exported and called from a
different module (i.e. reached after crossing an `import` boundary), throws
`ReferenceError: X is not defined` when its body reads a plain top-level
`const`/`let` declared in its *own* defining module — even though:

- the same binding reads correctly at that module's own top level, and
- the same function, called **locally** within its own module, reads it fine.

The distinguishing factor isn't "local vs. cross-module call" as such — it's
whether the defining module is the **entry module** or a **dependency**
loaded via the module loader. The identical file, run as the entry script,
works; imported as a dependency, it doesn't.

This is likely the single biggest blocker to running real, unmodified npm
ESM packages — "export a function that closes over module-level state,
call it from an importer" is close to the single most common pattern in
real-world JS.

## Repro (bare minimum, reproduces in vanilla `paserati` CLI, no host involved)

```js
// lib.js
export const X = true;
export function f() {
  if (X) return "yes";
  return "no";
}

// main.js
import { f, X } from "./lib.js";
console.log("X=", X, "f()=", f());
```

```
$ paserati --no-typecheck main.js
ReferenceError: X is not defined
    at f (<f>:3:1)
    at <script> (<script>:2:1)
```

Reading `X` directly (`import { X } from "./lib.js"; console.log(X)`, no
call through `f`) works fine and prints `true`. Calling `f()` from *within*
`lib.js` itself (append a call after the declarations, run `lib.js` as the
entry file) also works fine and prints `"yes"`.

Reproduces identically with:
- `export`ed or module-private (non-exported) `X`
- `const` or `let`
- `f` declared before or after `X` in source order
- `.js` (`--no-typecheck`, skip-typecheck path) **and** `.ts` (full
  typechecked path) — this is not specific to skip-typecheck

## Diagnosis so far

`OpGetGlobal` (`pkg/vm/vm.go` ~line 13360) reads a **numeric index baked in
at compile time** from the shared unified heap (`vm.heap.Get(int(globalIdx))`)
— not a name-based/module-relative lookup. When the slot doesn't exist, it
throws exactly this `ReferenceError` (`"%s is not defined"`, using
`heap.GetNameByIndex` only to recover a display name).

Confirmed the module's own top-level code *does* successfully assign `X`
into the heap (a `console.log` of `X` at that module's own top level,
[...]

PR #458, opened 2026-09-13
fix(compiler): export var now hoists correctly when compiled as an imported module

### LINKED ISSUE #451: compiler: a top-level function that forward-references a var declared later in the same file throws "is not defined" when that file is loaded as an imported module (works fine as the entry script)
Found continuing the zod investigation (docs/real-node-plan.md Round 104/106/107 in noderati) past `#432`/`#433` (both fixed and confirmed): real `zod@3.25.76`'s `v3/types.js` now parses and its full API loads, but calling `z.number()` (or `z.object`, `z.string`, etc. - anything with a `.create()` static) throws:

```
ReferenceError: ZodFirstPartyTypeKind is not defined
```

Zod's real source is exactly this shape - a class's static `.create` (an arrow function assigned after the class body) referencing a TS-compiled `enum` (`export var ZodFirstPartyTypeKind; (function(K){...})(ZodFirstPartyTypeKind || (ZodFirstPartyTypeKind = {}));`) declared much later in the same file - completely valid, ordinary `var` hoisting.

## Minimal, two-file repro

`lib.mjs`:
```js
function makeThing() {
  return Kind.Number;
}

export var Kind;
(function (K) {
  K["Number"] = "number";
})(Kind || (Kind = {}));

export { makeThing };
```
`main.mjs`:
```js
import { makeThing } from "./lib.mjs";
console.log(makeThing());
```

Real Node: `number`.

Paserati: running `lib.mjs` **directly as the entry script** (`paserati lib.mjs` with an added `console.log(makeThing())` at the bottom) **works correctly** - `Kind` resolves fine, matching real Node.

Running `main.mjs` (which imports `lib.mjs` as a dependency and calls the same function from there) **fails**:
```
ReferenceError: Kind is not defined
    at makeThing (<makeThing>:2:1)
```

## Narrowed precisely

- **Not a hoisting/hoisting-order bug in the general sense** - the identical file, run as the entry script, hoists and resolves `Kind` correctly even though `makeThing` is defined textually before `export var Kind;`.
- **Specific to the file being loaded as an *imported* module rather than the entry script.** Confirmed by running the exact same file both ways.
- **Not specific to cross-module invocation** either, in the sense of "caller in a different file" - even a synchronous call to `makeThing()` written at the *end of `lib.mjs` itself* (same file, same module) fails once that file is loaded via `import` from elsewhere, while it succeeds when that identical file is the entry script.
- **Reordering fixes it**: moving the `export var Kind; (function(K){...})(...)` block *before* `function makeThing() {...}` in the source makes the imported-module case work correctly too.

This points at the module-compilation path specifically (as opposed to the entry-script compilation path) not correctly pre-registering top-level `var`/function-declaration hoisting before compiling earlier statement bodies that reference them as free variables - i.e. a single-pass-in-textual-order compile for modules, vs. a hoisting pre-pass for the entry script.

## Real-world impact

This is exactly the shape real `zod`'s `v3/types.js` uses for essentially every schema builder (`z.string`, `z.number`, `z.boolean`, `z.date`, `z.bigint`, ... - all of them reference `ZodFirstPartyTypeKind`, declared at the very end of a ~3700-line file), so it currently blocks calling any of zod's real API once imported normally (`import { z } from "zod"`).

nooga:
## Summary

`collectVarDeclarations` (the var-hoisting name collector, used both to pre-register top-level names before hoisted functions compile and to emit the `undefined`-init/`MarkVarGlobal` hoist step) had no case for `*parser.ExportNamedDeclaration`, so `export var X;` was invisible to it - `X` never made it into `varNames` or `moduleBindings.TopLevelDeclNames`.

This was latent rather than immediately visible because `moduleGlobalKey` is a no-op for the persistent session compiler used to run a file directly as the entry script, so every reference to `X` (real declaration and forward reference alike) collapsed onto the same bare heap key by accident. Once the same file is compiled as an imported module (`loadedViaModuleLoader`), `moduleGlobalKey` namespaces top-level bindings by module path - and since `X` was missing from `TopLevelDeclNames`, a forward reference from a hoisted-and-compiled-early function fell through to the bare (unnamespaced) global key via `unresolvedGlobalKey`, while the actual declaration, compiled in sequence later, wrote to the namespaced key. Two different heap slots for the same binding → `ReferenceError: X is not defined`.

`collectLetConstDeclarations` already unwraps `ExportNamedDeclaration` for `export let`/`export const` (added for #117); this adds the identical case to `collectVarDeclarations` - the `var`-specific counterpart to #117.

**Real-world impact**: this is exactly the shape zod's `v3/types.js` uses for every schema builder (each references `ZodFirstPartyTypeKind`, an `export var` enum-shaped declaration near the end of the file) - blocking `import { z } from "zod"` entirely once loaded as a real module dependency rather than as an entry script.

## Testing
- New regression tests: `tests/scripts/module_export_var_forward_ref.ts` (+ helper)
- Full smoke suite green

## Note
The compiler-level bug (the reported symptom) is fixed. A related, separate checker-level gap remains: with type checking ON (the default), the same forward-reference shape still fails with `TS2304 Cannot find name 'Kind'` - reproduces on unmodified `main` too, so not a regression from this change. Flagged separately as a follow-up.

Fixes #451


nooga:
Superseded by #463, which contains this exact compiler-side fix (the same \`case *parser.ExportNamedDeclaration:\` added to \`collectVarDeclarations\`) plus the companion checker-level fix for the same forward-reference shape at type-check time. Closing this one unmerged in favor of #463 to avoid landing the identical change twice (and an add/add conflict on the shared test files). #451 is now closed via #463.

PR #510, opened 2026-09-22
Fix two stale smoke test expectations

nooga:
## Summary

Two smoke tests (`TestScripts`) have been failing consistently, unrelated to any recent functional changes. Root-caused both:

**`exception_pattern_bc_native_native.ts`**

The expected `JSON.parse` error message was accidentally truncated in commit `0bd49d8c` (an unrelated error-handling change) from `invalid character 'i' looking for beginning of object key string` down to just `invalid character 'i'`. The message itself comes verbatim from Go's `encoding/json` (`json_init.go` uses `err.Error()` directly) — not from paserati's own logic. Restored to match the current message (`invalid character 'i' looking for beginning of value`).

**`regexp_unicode_id_start_property.ts`**

Asserted `\p{ID_Continue}` should *not* match ZWNJ (U+200C), with a comment claiming "ZWNJ is Cf, not ID_Continue." Checked against the authoritative source, `unicode.org/Public/UCD/latest/ucd/DerivedCoreProperties.txt`:

```
200C..200D    ; ID_Continue # Cf   [2] ZERO WIDTH NON-JOINER..ZERO WIDTH JOINER
```

ZWNJ/ZWJ were added to `Other_ID_Continue` (and thus `ID_Continue`) specifically for identifier compatibility (e.g. Persian/Arabic). paserati's regex engine matching it is the spec-correct behavior; the test's assumption was wrong. Flipped the assertion and corrected the explanatory comment.

Neither of these is a functional regression in paserati — both predate this PR and are unrelated to it.

## Testing

- `go test ./tests -run TestScripts` — fully green.


nooga:
Closing — this PR's premise was wrong. I diagnosed both test failures using a local Go 1.27.1 toolchain, but this module pins `go 1.26.0` (go.mod), which is what CI actually installs. Both `encoding/json`'s error wording and the `unicode` package's `Other_ID_Continue` table (Go bundles a specific Unicode Character Database snapshot per release) differ between those two Go versions. Verified directly against `GOTOOLCHAIN=go1.26.0`: both original test expectations were already correct for the project's actual pinned toolchain. Pushed a revert; this branch is now a no-op against `main`.

Sorry for the noise — should have built/tested against the pinned toolchain from the start rather than whatever `go` happened to resolve to locally.

